Issue #44, opened 2022-08-15
Investigate what type of messagebox search requests that are slow


SandGrainOne:
### Description

There are a few search requests against the message box API that ends in timeout. We need to investigate what type of requests they are and decide to do to improve the performance.

In this issue we need to retrieve log data from AI and create an overview over parameter use like the number of parties.

### Additional Information
Slow requests last 2 months

```
7/31/2022, 8:42:13.486 PM
7/31/2022, 8:42:02.592 PM
7/31/2022, 8:40:58.575 PM
7/31/2022, 8:40:47.634 PM
7/24/2022, 6:32:31.452 AM
7/9/2022, 5:18:22.215 AM
7/9/2022, 1:33:22.314 AM
6/26/2022, 6:05:04.790 PM
```

### Tasks

- [x] Prepare telemetry to better support performance tests
- [ ] Analyze request bodies for slow requests
- [ ] Set up plan for performance tests

```
requests
| where url contains "sbl/instances" 
| where duration > 1000
| where cloud_RoleName == "platform-storage"
| order by timestamp desc
| project duration, customDimensions
```

### Acceptance Criterias

- [ ] A finished overview over the type of search requests that are too slow.
- [ ] How many single party requests and how many cross party searches during a week.
- [ ] Attempt to identify the type of user-party relationships.
- [ ] Is single party search slow in general or are the aspect with the slow searches? Both

acn-sbuad:
Ice box while we wait for telemetry to be available in production.

OddArneSaetervik:
Denne avventer vi avklaring og/eller tilbakekomst av ytelsesressurs

acn-sbuad:
Closing issue. As we are moving to postgresql performance on old infrastructure is no longer relevant.

PR #49, opened 2022-08-22
added ai filter to retrieve request body

acn-sbuad:
## Description
Adding request body as custom property of telemetry of slow requests. 

Should we log all request bodies for a shorter time period? 

## Related Issue(s)
- #44

## Verification
- [x] All tests run green


src/Storage/Filters/RequestBodyTelemetryFilter.cs
```diff
@@ -0,0 +1,55 @@
...
+                request.Properties.Add("queryModel", content);
+            }
+
+            Next.Process(item);
+        }
+
+        private string GetContentString(Stream content)
+        {
+            var reader = new StreamReader(content);
+            reader.BaseStream.Seek(0, SeekOrigin.Begin);
```
SandGrainOne, line 51: Would it be possible to create a test for this? Default behavior of the request stream is that it is none-reversible. We ran into issues with this before. I think there exists a property that enables seeking like this.

acn-sbuad: Good point. Changed implementation to get the telemetry directly in the controller. This will result in logging the request body for all requests for a while. Not sure what is a higher cost logging all bodies or including a stopwatch to only retrieve the bodies of requests that run long. 
This approach results in the fewest number of modifications to the code

src/Storage/Controllers/MessageboxInstancesController.cs
```diff
@@ -569,5 +573,16 @@ private async Task<ActionResult> ProcessQueryResponse(InstanceQueryResponse quer
 
             return Ok(authorizedInstances);
         }
+
+        private void AddQueryModelToTelemetry(MessageBoxQueryModel queryModel)
+        {
+            if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT").Equals("Development", StringComparison.InvariantCultureIgnoreCase))
```
SandGrainOne, line 579: Any reason for why we're excluding dev? Does it fail? Could we instead handle by checking the object returned by `Get<RequestTelemetry>()`?

acn-sbuad: Can change it to check the object. We don't have the connection to AI when running locally.  We need to have some sort of check to avoid an exception being thrown.

SandGrainOne (approved):
🎖️

PR #91, opened 2022-11-03
Feature/93 performance storage

HenningNormann:
Performance tuning

## Description
Avoid fetching data elements when getting instances for the message box

## Related Issue(s)
- #{93}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Repository/InstanceRepository.cs
+++ b/src/Storage/Repository/InstanceRepository.cs
@@ -82,7 +82,8 @@ namespace Altinn.Platform.Storage.Repository
         public async Task<InstanceQueryResponse> GetInstancesFromQuery(
             Dictionary<string, StringValues> queryParams,
             string continuationToken,
-            int size)
+            int size,
+            bool includeDataelements = true)
         {
             InstanceQueryResponse queryResponse = new InstanceQueryResponse
             {
@@ -120,7 +121,7 @@ namespace Altinn.Platform.Storage.Repository
                     }
 
                     List<Instance> instances = feedResponse.ToList();
-                    await PostProcess(instances);
+                    await PostProcess(instances, includeDataelements);
                     queryResponse.Instances.AddRange(instances);
                     queryResponse.Count += instances.Count;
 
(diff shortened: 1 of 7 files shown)


src/Storage/Repository/InstanceRepository.cs
```diff
@@ -564,13 +565,17 @@ private static void PreProcess(Instance instance)
...
             string instanceId = $"{instance.InstanceOwner.PartyId}/{instance.Id}";
 
             instance.Id = instanceId;
-            instance.Data = await _dataRepository.ReadAll(instanceGuid);
+            if (includeDataelements)
+            {
+                instance.Data = await _dataRepository.ReadAll(instanceGuid);
+            }
+
             if (instance.Data != null && instance.Data.Any())
```
acn-sbuad, line 579: Her er det vi tok opp i daily. Dette med at readStatus og lastChanged defineres basert på å sjekke instance og dataElements.

Men slik koden ser ut nå så vil jeg foreslå at sjekken på linje 579 flyttes inn i `includeDataElements`-delen. For den vil alltid evalueres til false hvis man ikke henter ut dataelementer.

HenningNormann: Dette skjønte jeg ikke. Hvis includeDataelementes==true, så blir jo alt riktig. Og hvis den er false, vil det jo uansett være umulig å få satt read og lastchanged riktig i forhold til å reflektere dataelementene.

SandGrainOne:
This pull request is on hold while we're considering issue #99 . We want the sorting order of instances in the message box to be fairly accurate and that depends on data from data elements.

src/Storage/Repository/InstanceRepository.cs
```diff
@@ -553,11 +552,13 @@ public async Task<Instance> GetOne(int instanceOwnerPartyId, Guid instanceGuid)
         /// <inheritdoc/>
         public async Task<Instance> Update(Instance item)
         {
+            List<DataElement> dataElements = item.Data;
             PreProcess(item);
 
             Instance instance = await Container.UpsertItemAsync<Instance>(item, new PartitionKey(item.InstanceOwner.PartyId));
 
-            await PostProcess(instance);
+            instance.Data = dataElements;
```
SandGrainOne, line 560: This is probably fine to do. I initially feared it would come back and bite us one day, but I can't see how that could happen at this point in time. The caller shouldn't expect changes to data elements when updating the instance.

b-brodie-digdir: Only case I can imagine in the future is if we were to start using server side triggers do more of the cross-document status synchronization (lastChanged timestamp for example). But that would require a larger refactoring to move dataElements into the same collection as instances, since server side triggers do not yet support cross-collection operations.

src/Storage/Repository/InstanceRepository.cs
b-brodie-digdir: This would of course require that dataElements be stored in the same container and use the same partition key. Would look something like this:
```
        public async Task<DataElement> Update(DataElement dataElement)
        {
            PartitionKey partitionKey = new PartitionKey(dataElement.InstanceGuid);

            TransactionalBatch batch = Container.CreateTransactionalBatch(partitionKey);

            batch.UpsertItem(dataElement);

            batch.PatchItem("instance_" + dataElement.InstanceGuid, new[]
            {
                PatchOperation.Set("/lastChanged", dataElement.LastChanged),
                PatchOperation.Set("/lastChangedBy", dataElement.LastChangedBy),
            });

            await batch.ExecuteAsync();

            return dataElement;
        }
```

SandGrainOne:
We'll revisit these changes and the ideas behind once we've completed the transition to PostgreSQL.

Issue #89, opened 2022-10-31
ProcessHistoryItem.Occurred is populated with created date from InstanceEvent


SandGrainOne:
### Description of the bug

During the mapping from instance events to `ProcessHistoryItem`, we use the `InstanceEvent` created date when populating the occured field. This will never be the same timestamp as the timestamps under the `ProcessInfo` property.

We need to look at the mapping to ensure we've picked the correct timestamp for each process history item.

### Steps To Reproduce

It's possible to reproduce the bug, but every single instance ever created can already be used to exemplify the issue.

### Additional Information

The issue was identified by brreg developers:
https://altinn.slack.com/archives/C02ESDSGPN1/p1667212783346829

### Suggested fix
We'll retrvive time stamp from ProcessInfo.Started and ProcessInfo.Ended for the two events respectively.

https://github.com/Altinn/altinn-storage/blob/main/src/Storage/Helpers/ProcessHelper.cs#L32

## Expected behaviour
The process history item occured property has the time stamp from the process event and not the instance event created property.

acn-sbuad:
Verified in AT

PR #100, opened 2022-11-29
Adjust mapping from ProcessState to ProcessHistoryItem

b-brodie-digdir:
When generating instance process history data, use ProcessInfo data when available.

## Description
The current implementation only refers to the instance Created timestamp, which should rather be used as a fallback for instances that do not include a process.

## Related Issue(s)
- #89 

## Verification
- [x] Code builds clean without any errors or warnings
- [x] All tests run green
- [x] Relevant automated test added (added 2 new test cases)
- [ ] Relevant stakeholders informed (check thread on Slack)

## Documentation
N/A

### DIFF
--- a/src/Storage/Helpers/ProcessHelper.cs
+++ b/src/Storage/Helpers/ProcessHelper.cs
@@ -29,9 +29,9 @@ namespace Altinn.Platform.Storage.Helpers
                         history.Add(
                             new ProcessHistoryItem
                             {
-                                Occured = instanceEvent.Created,
+                                Occured = instanceEvent?.ProcessInfo?.Started ?? instanceEvent.Created,
                                 EventType = instanceEvent.EventType,
-                                ElementId = instanceEvent.ProcessInfo.StartEvent,
+                                ElementId = instanceEvent?.ProcessInfo?.StartEvent,
                                 PerformedBy = GetPerformedBy(instanceEvent.User)
                             });
                         break;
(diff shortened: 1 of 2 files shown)


src/Storage/Helpers/ProcessHelper.cs
```diff
@@ -29,19 +29,19 @@ public static List<ProcessHistoryItem> MapInstanceEventsToProcessHistory(List<In
                         history.Add(
                             new ProcessHistoryItem
                             {
-                                Occured = instanceEvent.Created,
+                                Occured = instanceEvent?.ProcessInfo?.Started ?? instanceEvent.Created,
                                 EventType = instanceEvent.EventType,
-                                ElementId = instanceEvent.ProcessInfo.StartEvent,
+                                ElementId = instanceEvent?.ProcessInfo?.StartEvent,
```
b-brodie-digdir, line 34: It seems as though not all instances have a ProcessInfo object, which would result in a null exception here, that is potentially being swallowed somewhere in the call stack.
Is it relevant to investigate this further?

acn-sbuad: Don't think this is relevant to investigate further no.The instance events without processinfo are those not related to process steps but rather to enduser actions such as read and write.  No need for the null check either. if it is of the type `process_StartEvent` we can guarantee it isn't null

acn-sbuad (approved):
🎈

Issue #124, opened 2023-01-04
Retrieve dataElements in parallel to improve performance


b-brodie-digdir:
### Description

As an alternative to #99, we can request all dataElements for multiple instances in a single query. 

Preliminary testing performed on a instanceOwner with more than 100 instances indicates a reduction from >7 seconds to <700ms.

PR #125, opened 2023-01-04
Retrieve dataElements in parallel to save approx 90% RU

b-brodie-digdir:
Instead of requesting dataElements individually for each instance, define a cross-partition query that retrieves dataElements for multiple instances in a single query.

## Related Issue(s)
- #124 

## Verification
- [ ] All tests run green

### DIFF
--- a/src/Storage/Repository/DataRepository.cs
+++ b/src/Storage/Repository/DataRepository.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


src/Storage/Repository/IDataRepository.cs
```diff
@@ -44,6 +44,13 @@ public interface IDataRepository
         /// <returns>list of data elements</returns>
         Task<List<DataElement>> ReadAll(Guid instanceGuid);
 
+        /// <summary>
+        /// Gets all data elements for a given instance
```
acn-sbuad, line 48:
```suggestion
        /// Gets all data elements for given instances
```

src/Storage/Repository/DataRepository.cs
```diff
@@ -159,6 +160,45 @@ public async Task<List<DataElement>> ReadAll(Guid instanceGuid)
...
+            {
+                MaxBufferedItemCount = 0,
+                MaxConcurrency = -1,
+
+                // Do not set PartitionKey as we are querying across all partitions
+            };
+
+            FeedIterator<DataElement> query = Container
+                .GetItemLinqQueryable<DataElement>(requestOptions: options)
+                .Where(x => instanceGuids.Contains(x.Id))
```
acn-sbuad, line 187: Id here would be the id of the dataElement rather than the instanceGuid, won't it? 
Thinking this should be 
`.Where(x => instanceGuids.Contains(x.InstanceGuid ))`

src/Storage/Repository/InstanceRepository.cs
```diff
@@ -587,9 +587,37 @@ private async Task PostProcess(Instance instance)
         /// <param name="instances">the list of instances</param>
         private async Task PostProcess(List<Instance> instances)
         {
-            foreach (Instance item in instances)
+            Dictionary<string, Instance> instanceMap = instances.ToDictionary(key => key.Id, instance => instance);
+            var instanceGuids = instances.Select(i => i.Id).ToList();
+
+            // PreProcess all instances
```
SandGrainOne, line 593: Should probably have been PostProcess, but I suggest removing the comment. 
```suggestion
```

SandGrainOne (approved):

acn-sbuad (changes requested):
Would like to verify this locally before merge

src/Storage/Repository/InstanceRepository.cs
```diff
@@ -587,9 +587,33 @@ private async Task PostProcess(Instance instance)
...
+
+            foreach (var instanceGuid in instanceGuids)
+            {
+                instanceMap[instanceGuid].Data = dataElements[instanceGuid];
+            }
+
+            foreach (Instance instance in instances)
             {
-                await PostProcess(item);
+                instance.Id = InstanceIdToCosmosId(instance.Id);
```
acn-sbuad, line 607: InstanceId should be changed FROM cosmosId to regularId.
Check logic on line 570.

b-brodie-digdir: It was the other way around of course.

acn-sbuad (approved):

PR #133, opened 2023-01-19
Add FileScanClient and business logic

SandGrainOne:
## Description
Add FileScanClient to enqueue file scan requests
Add DataService to hold the (very limited) business logic.
Update Program to register new services

## Related Issue(s)
- https://github.com/Altinn/altinn-file-scan/issues/15


src/Storage/Controllers/DataController.cs
```diff
@@ -287,6 +294,8 @@ public async Task<ActionResult<DataElementList>> GetMany(int instanceOwnerPartyI
 
             DataElement dataElement = await _dataRepository.Create(newData);
             dataElement.SetPlatformSelfLinks(_storageBaseAndHost, instanceOwnerPartyId);
+            
+            await _dataService.PerformFileScan(dataTypeDefinition, dataElement, CancellationToken.None);
```
acn-sbuad, line 298: I think we need this for the PUT endpoint as well, in those cases they PUT on an existing data object

src/Storage/appsettings.json
```diff
@@ -32,5 +32,9 @@
   },
   "PepSettings": {
     "PdpDecisionCachingTimeout": 5
+  },
+  "FileScanQueueSettings": {
+    "FileScanQueueName": "file-scan-inbound",
+    "ConnectionString": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1"
```
acn-sbuad, line 38: I don't think we have a task for it, but this connection string should be stored in KV during generation of the storage account. Perhaps Consider agreeing on naming of the property with @tba76 so the secret is only stored once, but used by both storage and the function? given that we are referencing the same resource

acn-sbuad: Connection string is persisted in key vault during provisioning as `QueueStorageSettings--ConnectionString`

src/Storage/Controllers/DataController.cs
SandGrainOne: Added as requested

src/Storage/appsettings.json
acn-sbuad: My bad. The function secret is defined seperate of what is loaded into key vault, so duplicating settings name from the function won't work. We have two options here
1) Name it `QueueStorageSettings` to match what is is KV today for this storage account
2) Update tf to duplicate the secret with the desired name `FileScanQueueSettings` in KV during provisioning.

Thinking of it, maybe a 3rd as well.. Read the QueueStorageSettings into the `FileScanQueueSettings` class when setting up the configuration in Program.cs. Although this would make it hard to follow the code and find the settings use in kv at a later point I guess

test/UnitTest/TestingControllers/DataControllerTests.cs
```diff
@@ -65,13 +66,34 @@ public async void Post_NewData_Ok()
...
+        }
+
+        [Fact]
+        public async void Post_NewDataThatRequiresFileScan_Ok()
+        {
+            // Arrange
+            string dataPathWithData = $"{_versionPrefix}/instances/1337/bc19107c-508f-48d9-bcd7-54ffec905306/data";
+            HttpContent content = new StringContent("This is a blob file");
+
+            Mock<IFileScanQueueClient> fileScanMock = new Mock<IFileScanQueueClient>();
```
acn-sbuad, line 79: any reasoning behind mocking the queueClient rather than the dataService that the controller has a dependency on? Wouldn't it be cleaner to mock the immediate dependency ?

src/Storage/Program.cs
```diff
@@ -242,11 +243,13 @@ void ConfigureServices(IServiceCollection services, IConfiguration config)
     services.AddSingleton<ISasTokenProvider, SasTokenProvider>();
     services.AddSingleton<IKeyVaultClientWrapper, KeyVaultClientWrapper>();
     services.AddSingleton<IPDP, PDPAppSI>();
+    services.AddSingleton<IFileScanQueueClient, FileScanQueueClient>();
```
acn-sbuad, line 246: We might be missing the line to adding the FileScanQueueSettings to the configuration?

src/Storage/appsettings.json
acn-sbuad: To clarify, the issue isn't the queue name, but the if for the settings section `FileScanQueueSettings` which we will need to override with a connection from kv.

test/UnitTest/TestingControllers/DataControllerTests.cs
SandGrainOne: These tests already include a full in memory http service with an almost complete middleware. In my opinion not very clean to begin with. 

Whenever I use the in memory http service I like to mock as little as possible, or as late as possible. This gives them "deeper" integrations.  This is different from unit tests were I like to mock immediate dependencies.

src/Storage/Program.cs
SandGrainOne: Yes, good catch.

acn-sbuad (approved):
🎈

src/Storage/Altinn.Platform.Storage.csproj
```diff
@@ -9,12 +9,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Altinn.Platform.Storage.Interface" Version="3.11.0" />
+    <PackageReference Include="Altinn.Platform.Storage.Interface" Version="3.14.0" />
```
acn-sbuad, line 12:
```suggestion
    <PackageReference Include="Altinn.Platform.Storage.Interface" Version="3.15.0" />
```

src/Storage/Controllers/DataController.cs
```diff
@@ -287,6 +296,8 @@ public async Task<ActionResult<DataElementList>> GetMany(int instanceOwnerPartyI
 
             DataElement dataElement = await _dataRepository.Create(newData);
             dataElement.SetPlatformSelfLinks(_storageBaseAndHost, instanceOwnerPartyId);
+
```
acn-sbuad, line 299: Should the logic to add the hash to the data element be included here somewhere ?

src/Storage/Program.cs
b-brodie-digdir: Resolved on line 195?

b-brodie-digdir (commented):
Looks good to me.

src/Storage/appsettings.json
SandGrainOne: I would have preferred option 2, but have now implemented 1.

PR #137, opened 2023-01-23
Add ContentHash to DataElement model, bump versions.

b-brodie-digdir:
One more adjustment to DataElement model, saving base64 encoded MD5 content hash value computed by Azure Blob Storage on successful upload.

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Storage.Interface/Models/DataElement.cs
+++ b/src/Storage.Interface/Models/DataElement.cs
@@ -66,7 +66,7 @@ namespace Altinn.Platform.Storage.Interface.Models
         public long Size { get; set; }
 
         /// <summary>
-        /// Gets or sets the computed MD5 hash value of the blob.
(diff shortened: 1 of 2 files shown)


SandGrainOne (approved):

src/Storage.Interface/Models/DataElement.cs
```diff
@@ -65,6 +65,12 @@ public class DataElement : ChangableElement
         [JsonProperty(PropertyName = "size")]
         public long Size { get; set; }
 
+        /// <summary>
+        /// Gets or sets the computed MD5 hash value of the blob.
```
ivarne, line 69:
```suggestion
        /// Gets or sets the computed MD5 hash value of the blob. (Base64 encoded string not the more common hex encoding)
```

PR #194, opened 2023-04-26
Add config for shadow fields

nkylstad:
## Description

Update to DataType model for configuration of shadow fields in applicationMetadata.

## Related Issue(s)
- https://github.com/Altinn/altinn-studio/issues/10108

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Storage.Interface/Models/ShadowFields.cs
+++ b/src/Storage.Interface/Models/ShadowFields.cs
@@ -1,4 +1,4 @@
- using Newtonsoft.Json;
+using Newtonsoft.Json;
 
 namespace Altinn.Platform.Storage.Interface.Models
 {
@@ -20,6 +20,5 @@ namespace Altinn.Platform.Storage.Interface.Models
         /// </summary>
         [JsonProperty(PropertyName = "saveToDataType")]
         public string SaveToDataType { get; set; }
-
     }
 }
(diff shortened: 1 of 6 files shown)


src/Storage.Interface/Models/ShadowFields.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "prefix")]
+        public string Prefix { get; set; }
+
+        /// <summary>
+        /// Gets or sets the data type to save filtered data (without shadow fields) to.
+        /// Optional. If not set, the containing data type will be updated.
+        /// </summary>
+        [JsonProperty(PropertyName = "saveToDataType")]
+        public string SaveToDataType { get; set; }
```
acn-sbuad, line 22: Wondering why we need a reference to the dataType withing the shadowField object if it is already a sub-property of a given data type?

nkylstad: The property is meant to specify the data type to save the shadow fields to - i.e. a different data type. If specified, we will use this dataType and save the updated data to that dataType. If not specified, we use the containing data type (the one that this config is set for)

nkylstad: Would something like `DataTypeForFilteredData` be clearer?

acn-sbuad (approved):
🎈

src/Storage.Interface/Models/ShadowFields.cs
acn-sbuad: I think we can leave it as is. Based on your

PR #196, opened 2023-05-02
PoC implement PostgreSQL repositories to replace Cosmos repositories

HenningNormann:
## Description
- The implementation of PgInstanceRepository.GetInstancesFromQuery is postponed because the parameter parsing logic is somewhat complicated and not very important for the overall PoC
- Internal sequence based columns are used as primary keys in all tables. This is normally best practice and has the best join performance.

## Related Issue(s)
- #191

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Repository/PgApplicationRepository.cs
```diff
@@ -0,0 +1,217 @@
...
+using Npgsql;
+using NpgsqlTypes;
+
+namespace Altinn.Platform.Storage.Repository
+{
+    /// <summary>
+    /// Handles applicationMetadata repository. Notice that the all methods should modify the Id attribute of the
+    /// Application, since cosmosDb fails if Id contains slashes '/'.
+    /// </summary>
+    public class PgApplicationRepository : IApplicationRepository, IHostedService
```
SandGrainOne, line 26: It's should be unnecessary to make the new repository classes be IHostedService.

SandGrainOne (commented):
First findings. Need to look at the PR some more.

src/Storage/Repository/PgApplicationRepository.cs
HenningNormann: I can't see how to avoid this as long as we need to be able to invoke both the old and the new repository classes. When the convertion is complete, we can remove the old classes and the IHostedService interface.

src/Storage/Program.cs
```diff
@@ -337,6 +348,27 @@ void Configure()
         app.UseExceptionHandler("/storage/api/v1/error");
     }
 
+    if (config.GetValue<bool>("PostgreSQLSettings:EnableDBConnection"))
```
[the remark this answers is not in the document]
HenningNormann: Done

src/Storage/appsettings.json
```diff
@@ -1,4 +1,12 @@
 {
+  "PostgreSQLSettings": {
+    "EnableDBConnection": "true",
```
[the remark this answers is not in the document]
HenningNormann: Done

SandGrainOne (commented):
Found a few more things to discuss

src/Storage/Repository/PgInstanceRepository.cs
```diff
@@ -0,0 +1,333 @@
...
+            else
+            {
+                pgcom.Parameters.AddWithValue(NpgsqlDbType.Integer, 0);
+            }
+
+            pgcom.Parameters.AddWithValue(NpgsqlDbType.Text, taskPredicate != null ? queryParams["process.currentTask"].First() : DBNull.Value);
+            pgcom.Parameters.AddWithValue(NpgsqlDbType.Text, orgPredicate != null ? queryParams["org"].First() : DBNull.Value);
+            pgcom.Parameters.AddWithValue(NpgsqlDbType.Text, appIdPredicate != null ? queryParams["appId"].First() : DBNull.Value);
+            pgcom.Parameters.AddWithValue(NpgsqlDbType.TimestampTz, lastChanged);
+            pgcom.Parameters.AddWithValue(NpgsqlDbType.TimestampTz, created);
```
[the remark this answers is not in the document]
HenningNormann: Yes, I know. That's also true for task, org and appid. What's your concern? lastChanged and created are not nullable in the client side code and don't need any kind of  DBNull.Value treatment. If the values are not present in the input, they will not be referenced in the sql - the same way as for task, org and appid.

The only way to completely avoid sending sql parameters that are not provided by the caller, is to have the sql completely  dynamic - without using bind variables. But that would be bad regarding performance.

src/Storage/Repository/PgInstanceRepository.cs
```diff
@@ -0,0 +1,333 @@
...
+
+            await using (NpgsqlDataReader reader = await pgcom.ExecuteReaderAsync())
+            {
+                long previousId = -1;
+                while (await reader.ReadAsync())
+                {
+                    id = reader.GetFieldValue<long>("id");
+                    if (id != previousId)
+                    {
+                        SetStatuses(instance);
```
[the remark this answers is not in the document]
HenningNormann: No, I think it's correct. When all the data elements are read for an instance, the aggregated status for the instance must be updated. The alternative would be to loop through all the instances in a separate loop, but why do that?

src/Storage/appsettings.json
```diff
@@ -1,4 +1,12 @@
 {
+  "PostgreSQLSettings": {
```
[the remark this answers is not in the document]
HenningNormann: Done

src/Storage/Controllers/InstancesController.cs
```diff
@@ -275,6 +275,35 @@ public async Task<ActionResult<Instance>> Get(int instanceOwnerPartyId, Guid ins
...
+        /// </summary>
+        /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+        /// <param name="instanceGuid">The id of the instance to retrieve.</param>
+        /// <returns>Authorization info.</returns>
+        [HttpGet("AuthInfo/{instanceOwnerPartyId:int}/{instanceGuid:guid}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [Produces("application/json")]
+        [ApiExplorerSettings(IgnoreApi = true)]
+        public async Task<ActionResult<(ProcessState Process, string AppId)>> GetForAuth(int instanceOwnerPartyId, Guid instanceGuid)
```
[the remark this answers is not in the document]
HenningNormann: - New object model done. Upgrade of version number and reference is not yet done.
- Conclusion from architecture meeting was that access token was not necessary
- Move to ProcessController not yet done

src/Storage/Program.cs
```diff
@@ -337,6 +348,27 @@ void Configure()
         app.UseExceptionHandler("/storage/api/v1/error");
     }
 
+    if (config.GetValue<bool>("PostgreSQLSettings:EnableDbConnection") && config.GetValue<bool>("generalSettings.UsePostgreSQL"))
```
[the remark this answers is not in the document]
HenningNormann: It's removed now.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -276,6 +276,35 @@ public async Task<ActionResult<Instance>> Get(int instanceOwnerPartyId, Guid ins
...
+        [Produces("application/json")]
+        [ApiExplorerSettings(IgnoreApi = true)]
+        public async Task<ActionResult<AuthInfo>> GetForAuth(int instanceOwnerPartyId, Guid instanceGuid)
+        {
+            try
+            {
+                (Instance instance, _) = await _instanceRepository.GetOne(instanceOwnerPartyId, instanceGuid, false);
+                if (instance.InstanceOwner.PartyId != instanceOwnerPartyId.ToString())
+                {
+                    throw new Exception("Party id not found");
```
SandGrainOne, line 297: Throwing a new exception for the sole purpose of breaking out of the code block should be avoided.

I can suggest two alternatives:
1. return NotFound().
2. Invert the if and adjust logic to match.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -276,6 +276,35 @@ public async Task<ActionResult<Instance>> Get(int instanceOwnerPartyId, Guid ins
...
+        /// </summary>
+        /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+        /// <param name="instanceGuid">The id of the instance to retrieve.</param>
+        /// <returns>Authorization info.</returns>
+        [HttpGet("AuthInfo/{instanceOwnerPartyId:int}/{instanceGuid:guid}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [Produces("application/json")]
+        [ApiExplorerSettings(IgnoreApi = true)]
+        public async Task<ActionResult<AuthInfo>> GetForAuth(int instanceOwnerPartyId, Guid instanceGuid)
```
SandGrainOne, line 290: We'll need to add the new type in a separate PR, merge and publish a new version of the package.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -276,6 +276,35 @@ public async Task<ActionResult<Instance>> Get(int instanceOwnerPartyId, Guid ins
             }
         }
 
+        /// <summary>
+        /// Gets process info relevant to authorization
+        /// </summary>
+        /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+        /// <param name="instanceGuid">The id of the instance to retrieve.</param>
+        /// <returns>Authorization info.</returns>
+        [HttpGet("AuthInfo/{instanceOwnerPartyId:int}/{instanceGuid:guid}")]
```
SandGrainOne, line 285: URLs are case insensitive in most cases, but I prefer that we define routes in lowercase.

src/Storage/Controllers/InstancesController.cs
HenningNormann: Done

src/Storage/Controllers/InstancesController.cs
HenningNormann: Done

src/Storage/Controllers/InstancesController.cs
HenningNormann: https://github.com/Altinn/altinn-storage/pull/233

SandGrainOne: We've now published version 3.23.0

test/UnitTest/TestingControllers/DataControllerUnitTests.cs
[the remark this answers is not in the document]
HenningNormann: Fixed now

src/Storage/appsettings.json
```diff
@@ -1,4 +1,11 @@
...
+    "WorkspacePath": "Migration",
+    "AdminConnectionString": "Host=localhost;Port=5432;Username=platform_storage_admin;Password={0};Database=storagedb",
+    "ConnectionString": "Host=localhost;Port=5432;Username=platform_storage;Password={0};Database=storagedb;Include Error Detail=True",
+    "StorageDbAdminPwd": "<get-from-key-vault>",
+    "StorageDbPwd": "<get-from-key-vault>"
```
acn-sbuad, line 7:
```suggestion
    "StorageDbPwd": "Password"
```

let's include default password to simplify local development. No KV involved then.

src/Storage/appsettings.json
```diff
@@ -1,4 +1,11 @@
 {
+  "PostgreSQLSettings": {
+    "WorkspacePath": "Migration",
+    "AdminConnectionString": "Host=localhost;Port=5432;Username=platform_storage_admin;Password={0};Database=storagedb",
+    "ConnectionString": "Host=localhost;Port=5432;Username=platform_storage;Password={0};Database=storagedb;Include Error Detail=True",
+    "StorageDbAdminPwd": "<get-from-key-vault>",
```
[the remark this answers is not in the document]
SandGrainOne: I would prefer if we started using UserSecrets instead of relying on correct settings in the appsettings files. At a minimum, move the correct value to the appsettings.Developmen.json file.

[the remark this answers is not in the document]
HenningNormann: Done

src/Storage/Repository/PgDataRepository.cs
```diff
@@ -0,0 +1,336 @@
...
+using NpgsqlTypes;
+
+namespace Altinn.Platform.Storage.Repository
+{
+    /// <summary>
+    /// Represents an implementation of <see cref="IDataRepository"/>.
+    /// </summary>
+    public class PgDataRepository : IDataRepository
+    {
+        private readonly string _insertSql = "call storage.insertDataelement ($1, $2, $3, $4)"; //"insert into storage.dataelements(instanceInternalId, instanceGuid, alternateId, element) VALUES ($1, $2, $3, $4);";
```
SandGrainOne, line 26:
```suggestion
         //"insert into storage.dataelements(instanceInternalId, instanceGuid, alternateId, element) VALUES ($1, $2, $3, $4);";
        private readonly string _insertSql = "call storage.insertDataelement ($1, $2, $3, $4)"; 
```
Move the comment to above the statement. use four / instead of two. See if that removes all these warnings. Repeat for the other statmements as well. You might be forced to add a blank line between statements.

src/Storage/Repository/PgApplicationRepository.cs
```diff
@@ -0,0 +1,198 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Npgsql;
+using NpgsqlTypes;
+
+namespace Altinn.Platform.Storage.Repository
+{
+    /// <summary>
+    /// Handles applicationMetadata repository. Notice that the all methods should modify the Id attribute of the
+    /// Application, since cosmosDb fails if Id contains slashes '/'.
```
SandGrainOne, line 24: This summary was copied over from the other implementation.

src/Storage/Repository/PgApplicationRepository.cs
```diff
@@ -0,0 +1,198 @@
...
+                    appTitles.Add(item.Id, titles.ToString());
+                }
+
+                _memoryCache.Set(_cacheKey, appTitles, _cacheEntryOptionsTitles);
+            }
+
+            return appTitles;
+        }
+
+        private static Application SetLegacyId(Application app)
```
SandGrainOne, line 178: Now that we no longer need to avoid / in the id because of limitations in CosmosDB, we could choose to store the value with / and avoid the headache of maintaining this special case.

src/Storage/Repository/PgApplicationRepository.cs
```diff
@@ -0,0 +1,198 @@
...
+            if (app.Id.StartsWith(app.Org + '-', StringComparison.OrdinalIgnoreCase))
+            {
+                string appIdWithoutOrg = app.Id[(app.Org.Length + 1)..];
+                app.Id = $"{app.Org}/{appIdWithoutOrg}";
+            }
+
+            return app;
+        }
+
+        private static void SetInternalId(Application app)
```
SandGrainOne, line 189: Now that we no longer need to avoid / in the id because of limitations in CosmosDB, we could choose to store the value with / and avoid the headache of maintaining this special case. We need something during migration though. Any thoughts on this?

src/Storage/Migration/v0.00/01-setup-tables.sql
```diff
@@ -0,0 +1,80 @@
...
+	created TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+	lastChanged TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+	org TEXT NOT NULL,
+	appId TEXT NOT NULL,
+	taskId TEXT NULL,
+	instance JSONB NOT NULL
+)
+TABLESPACE pg_default;
+
+CREATE TABLE IF NOT EXISTS storage.dataElements
```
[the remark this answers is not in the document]
HenningNormann: Such a change will also influence the id in the text resource. It's not recommended to do this now.

src/Storage/Repository/PgApplicationRepository.cs
HenningNormann: Such a change will also influence the id in the text resource. It's not recommended to do this now.

src/Storage/Repository/PgApplicationRepository.cs
HenningNormann: Such a change will also influence the id in the text resource. It's not recommended to do this now.

src/Storage/Repository/PgDataRepository.cs
HenningNormann: Done

src/Storage/Repository/PgApplicationRepository.cs
HenningNormann: Corrected

src/Storage/Repository/PgApplicationRepository.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <inheritdoc/>
+        public async Task<Dictionary<string, string>> GetAllAppTitles()
+        {
+            if (!_memoryCache.TryGetValue(_cacheKey, out Dictionary<string, string> appTitles))
+            {
+                appTitles = new Dictionary<string, string>();
+                foreach (Application item in await FindAll())
+                {
+                    StringBuilder titles = new();
+                    if (item?.Title?.Values != null)
```
SandGrainOne, line 160: The "bug" being reported by SonarCloud might be linked to the question mark right after item. Try to remove it. It should be impossible for item to be null here. 
```suggestion
                    if (item.Title?.Values != null)
```

src/Storage/appsettings.Development.json
```diff
@@ -1,4 +1,11 @@
 {
+  "PostgreSQLSettings": {
```
SandGrainOne, line 2: Looks like the settings reader is case insensitive. The code looks for section `PostgreSqlSettings`.

src/Storage/Repository/PgApplicationRepository.cs
HenningNormann: Done

src/Storage/appsettings.Development.json
HenningNormann: Done

PR #176, opened 2023-03-16
Set the AllowNewCopy property for an archived instance

SandGrainOne:
## Description
Set the AllowNewCopy property for an archived instance. AllowNewCopy is the be set for archived instances for apps where the copy feature is enabled and the user have the necessary rights.

## Related Issue(s)
- https://github.com/Altinn/altinn-platform/issues/355

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #206, opened 2023-05-25
Add new controller with endpoints for lock/unlock DataElements

tjololo:
## Description
Locking data is currently done through the `PUT /data/{guid}` endpoint. This has the undesired effect that trying to lock already locked dataelements result in a 409 Conflict respons. Locking data should be an idempotent operation so locking already locked data is ok.
Also introducing a unlock endpoint where users with the action unlock granted can unlock data. This is to make it possible for a reviewer of an instance to "reject" the data and "send" the instance back in the process and unlock the data even if he/she is not allowed to update/write data 

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation


src/Storage/Controllers/DataLockController.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    /// <summary>
+    /// Locks a data element
+    /// </summary>
+    /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+    /// <param name="instanceGuid">The id of the instance that the data element is associated with.</param>
+    /// <param name="dataGuid">The id of the data element to delete.</param>
+    /// <returns></returns>
+    [Authorize(Policy = AuthzConstants.POLICY_INSTANCE_WRITE)]
+    [HttpPut]
```
SandGrainOne, line 50: If "lock" is the REST resource then this should be a POST.

tjololo: Agreed, was so focused on the dataelemt part that I forgot that we actually are creating/deleting a lock.

tjololo: As we discussed, left the method as PUT but it returns 200 if already locked and 201 if not previously locked

SandGrainOne: We decided to keep PUT as it is okay to use it to create a resource and is assumed to be idempotent.

src/Storage/Controllers/DataLockController.cs
```diff
@@ -0,0 +1,142 @@
...
+        _authorizationService = authorizationService;
+    }
+
+    /// <summary>
+    /// Locks a data element
+    /// </summary>
+    /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+    /// <param name="instanceGuid">The id of the instance that the data element is associated with.</param>
+    /// <param name="dataGuid">The id of the data element to delete.</param>
+    /// <returns></returns>
```
SandGrainOne, line 49: Document return value

src/Storage/Controllers/DataLockController.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Unlocks a data element
+    /// </summary>
+    /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+    /// <param name="instanceGuid">The id of the instance that the data element is associated with.</param>
+    /// <param name="dataGuid">The id of the data element to delete.</param>
+    /// <returns></returns>
```
SandGrainOne, line 94: Document return value

src/Storage/Controllers/DataLockController.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            { "/locked", true }
+        };
+        
+        try
+        {
+            DataElement updatedDataElement = await _dataRepository.Update(instanceGuid, dataGuid, propertyList);
+            return Created(updatedDataElement.Id, updatedDataElement);
+        }
+        catch (CosmosException e)
```
SandGrainOne, line 82: We're in the process of changing from Cosmos to PostgreSQL. It's important that nothing related to Cosmos appear outside the Repository classes.

src/Storage/Authorization/AuthorizationService.cs
```diff
@@ -152,6 +149,48 @@ public async Task<bool> AuthorizeInstanceAction(Instance instance, string action
...
+            }
+            else
+            {
+                Guid instanceGuid = Guid.Parse(instance.Id.Split('/')[1]);
+                request = DecisionHelper.CreateDecisionRequest(org, app, user, actions[0], instanceOwnerPartyId, instanceGuid, task);
+            }
+
+            for (int i = 1; i < actions.Count; i++)
+            {
+                request.Request.Action.Add(DecisionHelper.CreateActionCategory(actions[i]));
```
SandGrainOne, line 177: This will probably not work. We would need to create individual DecisionRequests for each action. We have examples of this in relation to MessageBox functionality. We might be able to use the CreateMultiDecisionRequest method almost as is.

src/Storage/Authorization/AuthorizationService.cs
```diff
@@ -152,6 +149,48 @@ public async Task<bool> AuthorizeInstanceAction(Instance instance, string action
             return authorized;
         }
 
+        /// <inheritdoc/>>
+        public async Task<bool> AuthorizeAnyOfInstanceActions(Instance instance, List<string> actions, string task = null)
```
SandGrainOne, line 153: You're trying to make the method I decided not to make when I worked on the AllowNewCopy change. See PR #176 .

I would sleep better if we used the `CreateMultiDecisionRequest` method to create the request.

tjololo: I had a look at that method, but as this is not making a single decision and not multiple I tested this solution, but might be the auth mock that is behaving different than the real auth

test/UnitTest/Mocks/Repository/DataRepositoryMock.cs
```diff
@@ -8,6 +8,7 @@
 using Altinn.Platform.Storage.Interface.Models;
 using Altinn.Platform.Storage.Repository;
 using Altinn.Platform.Storage.UnitTest.Utils;
+using Microsoft.Azure.Cosmos;
```
SandGrainOne, line 11: Unnecessary using

SandGrainOne (approved):

Issue #213, opened 2023-05-31
Implement presentation of instances in signing task in messagebox


tba76:
## Considerations
- Do we need to define element.RoleRequirement  for the reporteeElemen to get the correct status? 
Yes - in storage we must check if user has rights to sign or just to read. Based on this a new propert(y/ies) should be included in the messageboxInstance class. 

### Tasks
- [x] Add signing as an action to the multidecision request for filtering msgbox instances
- [x] extend messagebox instance and instance event type models
  - [x] push new storage.interface
- [x] Implement required logic to handle signing when authorizing instances
- [x] Implement required logic to handle retrieval of signing related instance events
  - [x] test in SBL
- [x] Confirm the name of the task to ensure that element is placed correctly in msgbox 
- [x] implement required changes in sbl for reportee element list and activity log
- [x] QA
- [x] manual test

### Acceptance Criterias
- [x] Activity log displays events of Signed and Sent to signing
- [x] Button should say "Gå til skjema" when element in sign task is expanded
- [x] Element you can read but not sign should say "Venter på andre"
- [x] Element you can sign should say "Til signering;"
- [x] Elements sorted based on date alongside elements in data task

PR #217, opened 2023-06-01
new instance event types for signature

acn-sbuad:
## Description

Extended InstanceEvents enum with two values for signing events.

## Related Issue(s)
- #213


tba76 (approved):
🍺

PR #220, opened 2023-06-05
Generate instance events for signing events

acn-sbuad:
## Description
Logic related to generation of instance events in relation to signing
New instance event service used in all controllers. 
new instance service used for signing
Generate sent to sign event in processs controller if process step is signing. 

image

## Related Issue(s)
- #221

## Verification
- [x] All tests run green

### DIFF
--- a/src/Storage/Controllers/SignController.cs
+++ b/src/Storage/Controllers/SignController.cs
@@ -1,7 +1,10 @@
 using System;
 using System.Threading.Tasks;
+
 using Altinn.Platform.Storage.Helpers;
 using Altinn.Platform.Storage.Interface.Models;
+using Altinn.Platform.Storage.Services;
+
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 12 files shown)


src/Storage/Controllers/SignController.cs
```diff
@@ -35,6 +41,7 @@ public SignController()
         [Produces("application/json")]
         public async Task<ActionResult> Sign([FromRoute] int instanceOwnerPartyId, [FromRoute] Guid instanceGuid, [FromBody] SignRequest signRequest)
         {
+            await _instanceService.CreateSignDocument(instanceOwnerPartyId, instanceGuid, signRequest);
```
SandGrainOne, line 44: Should we return the signing document?

acn-sbuad: We discussed this friely, but didn't conclude with anything. Service is still being developed by Thomas so output can be updated. Simply cherry picked the files I needed from his dev branch to be able to complete implementation of the event logic

test/UnitTest/TestingControllers/ProcessControllerTest.cs
```diff
@@ -269,13 +268,55 @@ public async void PutProcess_EndProcess_EnsureArchivedStateIsSet()
             Assert.True(actual.Status.IsArchived);
         }
 
-        private HttpClient GetTestClient(IInstanceRepository instanceRepository = null)
+        /// <summary>
+        /// Test case: Uses want to go back to a earlier state
```
SandGrainOne, line 272: Description doesn't match test method name

Issue #223, opened 2023-06-06
Support references on data elements


acn-sbuad:
### Description

Related to: altinn/app-lib-dotnet#137
The application needs a way to know which data elements are generated as a result of a particular dataElement to be able to unlock/delete them if process is rolled back.

### Tasks
- [x] Update storage interface
- [x] Update controller endpoints in storage to accept reference info
- [x] QA
- [x] Duplicate logic in localtest
- [x] Manual test
- [x] Automated tests

### Acceptance Criterias
- [x] DataElement references are persisted on the dataElement object
- [x]  Value, ValueType and ReferenceType are set for each reference generated

PR #224, opened 2023-06-06
Support references on dataElements

acn-sbuad:
## Related Issue(s)
- #223 

## Verification
- [x] All tests run green

### DIFF
--- a/src/Storage/Controllers/DataController.cs
+++ b/src/Storage/Controllers/DataController.cs
@@ -236,6 +236,7 @@ namespace Altinn.Platform.Storage.Controllers
         /// <param name="instanceGuid">The id of the instance that the data element is associated with.</param>
         /// <param name="dataType">The data type identifier for the data being uploaded.</param>
         /// <param name="refs">An optional array of data element references.</param>
+        /// <param name="generatedFromIds">An optional array of generated from references containing data element ids.</param>
         /// <returns>The metadata of the new data element.</returns>
         [Authorize(Policy = AuthzConstants.POLICY_INSTANCE_WRITE)]
         [HttpPost("data")]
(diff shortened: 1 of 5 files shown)


acn-sbuad:
Logic related to the code smell is being considered moved into a new service in another PR, so leaving it as is for now.

src/Storage/Helpers/DataElementHelper.cs
```diff
@@ -38,13 +40,30 @@ public static DataElement CreateDataElement(string dataType, List<Guid> refs, In
                 LastChanged = creationTime,
                 Size = fileSize,
                 Refs = refs,
-            };            
+            };
+
+            if (generatedFromId != null && generatedFromId.Any())
```
SandGrainOne, line 45: Use the List.Count before IEnumeration.Any().

SandGrainOne: I keep forgetting that Any is smart and can detect that the underlying type is a List. Your choice.

SandGrainOne: Last comment. It's technically unnecessary to check for elements before a foreach. It's built in.

SandGrainOne (approved):

src/Storage/Helpers/DataElementHelper.cs
acn-sbuad: but we avoid instantiating the references list if we don't have any references to include, so I guess the check still makes sense?

src/Storage/Controllers/DataController.cs
```diff
@@ -445,6 +454,7 @@ public async Task<ActionResult<DataElement>> OverwriteData(int instanceOwnerPart
             {
                 { "/locked", dataElement.Locked },
                 { "/refs", dataElement.Refs },
+                { "/references", dataElement.References },
```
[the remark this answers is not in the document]
acn-sbuad: Limit is 10, så we still have 3-4 to go. But good point, I was not aware of the limitation.

src/Storage/Helpers/DataElementHelper.cs
SandGrainOne: Yeah, keep it.

With that being said there is another side to that discusion we should take one day. The nullable discussion. The rules say that a List reference should never be null. Just nitpicking. 😄

PR #232, opened 2023-06-12
Feature/change referencetype to task

tjololo:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Helpers/DataElementHelper.cs
+++ b/src/Storage/Helpers/DataElementHelper.cs
@@ -1,8 +1,7 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
-
 using Altinn.Platform.Storage.Interface.Models;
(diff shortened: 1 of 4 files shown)


src/Storage/Helpers/DataElementHelper.cs
```diff
@@ -1,8 +1,7 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
-
 using Altinn.Platform.Storage.Interface.Models;
+using Microsoft.IdentityModel.Tokens;
```
acn-sbuad, line 4: Doesn't seem this using is required for the new functionality?

tjololo: I used it to check if generatedFromTask is null or empty, but can rewrite check to use string.IsNullOrEmpty(generatedFromTask) instead :)

PR #251, opened 2023-08-24
return bad request if empty blob persisted

acn-sbuad:
## Description

There is a number of empty blbs in blob storage. As the metadatasays the size of the dataelement is 0, we should be able to reject these uploads and remove them from blob storage and indicate to the caller that an error occured.

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green


tba76 (approved):
🍺

PR #254, opened 2023-08-30
Bugfixes storage postgres interface

HenningNormann:
Fixes since last review. The migration files are kept in the same version to make them more readable until prod convertion. The database is updated manually to support this.

## Description
- Moved all filter logic in GetInstancesFromQuery into the database
- Added index for partyid lastchanged
- Changed delete procedures to functions to be able to return row count

## Related Issue(s)
- #191 

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Migration/v0.00/01-setup-tables.sql
```diff
@@ -74,7 +74,10 @@ CREATE INDEX IF NOT EXISTS instances_partyid ON storage.instances(partyId);
 CREATE INDEX IF NOT EXISTS instances_appid ON storage.instances(appId);
 CREATE INDEX IF NOT EXISTS instances_appid_taskId ON storage.instances(appId, taskId);
 CREATE INDEX IF NOT EXISTS instances_appid_lastchanged ON storage.instances(appId, lastChanged);
+CREATE INDEX IF NOT EXISTS instances_partyid_lastchanged ON storage.instances(partyId, lastChanged);
```
acn-sbuad, line 77: editing an existing migration version will require manual intervention in those environments where db has been deployed previously

src/Storage/Services/IDataService.cs
```diff
@@ -40,6 +40,7 @@ public interface IDataService
         /// <param name="org">The application owner id.</param>
         /// <param name="stream">Data to be written to blob storage.</param>
         /// <param name="dataElement">The data element to insert.</param>
-        Task UploadDataAndCreateDataElement(string org, Stream stream, DataElement dataElement);
+        /// <param name="internalId">The internal id of the data element to insert.</param>
```
acn-sbuad, line 43: do we need a better description of this concept of internal id? db id might be more descriptive? 
Having consequences of database limitations exposed in the service layer isn't optimal, but I guess we don't have any other options here?

src/Storage/Services/DataService.cs
```diff
@@ -71,12 +71,12 @@ public async Task<(string FileHash, ServiceError ServiceError)> GenerateSha256Ha
         }
 
         /// <inheritdoc/>
-        public async Task UploadDataAndCreateDataElement(string org, Stream stream, DataElement dataElement)
+        public async Task UploadDataAndCreateDataElement(string org, Stream stream, DataElement dataElement, long internalId)
```
acn-sbuad, line 74: unclear which objects internalId we are reffering to here.

src/Storage/Migration/v0.00/01-setup-tables.sql
HenningNormann: Ref. comment in PR. This is done deliberately to make it easier to read, maintain and compare db changes until conversion. The conversion requires manual operations anyway.

src/Storage/Services/IDataService.cs
HenningNormann: I see your concern, but I think the major problem here are all the different names of the existing keys – id, instanceid, instanceguid etc. To me this is very confusing, and I think it’s more important to change these names after cosmos is removed. In Altinn II the naming convention is id for primary key and <table>id for foreign key. I think this is a good convention, but it’s impossible to use now without changing the existing key names.

src/Storage/Services/DataService.cs
HenningNormann: Have renamed to instanceInternalId now. The preferred solution would be to redesign the interfaces so that the id would be internal to the repository methods, but I wouldn’t like to introduce such a change in the first release.

src/Storage/Repository/InstanceQueryHelper.cs
```diff
@@ -3,8 +3,11 @@
 using System.Linq;
 using Altinn.Platform.Storage.Helpers;
 using Altinn.Platform.Storage.Interface.Models;
+using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Extensions.Primitives;
+using Npgsql;
+using NpgsqlTypes;
```
SandGrainOne, line 10: None of the three new namespaces are actually used in the code.

src/Storage/Migration/v0.00/01-setup-tables.sql
SandGrainOne: I share the concern and are afraid we might forget something along the way if too much is to be handled manually.

HenningNormann: Let's discuss the conversion approach.

src/Storage/Controllers/CleanupController.cs
```diff
@@ -0,0 +1,174 @@
...
+            _dataRepository = dataRepository;
+            _instanceEventRepository = instanceEventRepository;
+            _logger = logger;
+        }
+
+        /// <summary>
+        /// Invoke periodic cleanup of instances
+        /// </summary>
+        /// <returns>?</returns>
+        [HttpGet("cleanupinstances")]
```
SandGrainOne, line 59: I want the endpoint(s) to be POST just out of principle.

src/Storage/Controllers/CleanupController.cs
```diff
@@ -0,0 +1,174 @@
...
+                stopwatch.Elapsed.TotalSeconds);
+
+            return Ok();
+        }
+
+        /// <summary>
+        /// Invoke periodic cleanup of data elements
+        /// </summary>
+        /// <returns>?</returns>
+        [HttpGet("cleanupdataelements")]
```
SandGrainOne, line 137: There is also the loadtest cleanup.

src/Storage/Repository/IInstanceRepository.cs
```diff
@@ -49,5 +49,17 @@ public interface IInstanceRepository
...
+        /// Gets hard deleted instances for cleanup
+        /// </summary>
+        /// <returns>Hard deleted instances</returns>
+        Task<List<Instance>> GetHardDeletedInstances();
+
+        /// <summary>
+        /// Gets hard deleted data elements for cleanup
+        /// </summary>
+        /// <returns>Hard deleted data elements</returns>
+        Task<List<DataElement>> GetHardDeletedElements();
```
SandGrainOne, line 63:
```suggestion
        Task<List<DataElement>> GetHardDeletedDataElements();
```

SandGrainOne:
There are failing tests.

src/Storage/Controllers/CleanupController.cs
HenningNormann: Added now

src/Storage/Controllers/CleanupController.cs
```diff
@@ -0,0 +1,236 @@
...
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ApiExplorerSettings(IgnoreApi = true)]
+        public async Task<ActionResult> CleanupInstancesForApp(string appId)
+        {
+            if (!_usePostgreSQL)
+            {
+                return Ok();
+            }
+
+            appId = appId.Contains('/') ? appId : appId.Replace(appId.Split('-').First() + '-', appId.Split('-').First() + '/');
```
SandGrainOne, line 107: I don't think we need to support appId on any format other than (org)/(appname).

src/Storage/Controllers/CleanupController.cs
```diff
@@ -0,0 +1,236 @@
...
+            }
+
+            stopwatch.Stop();
+            _logger.LogInformation(
+                $"NightlyCleanupDataElements // Run // {successfullyDeleted} of {dataElements.Count} data elements deleted in {stopwatch.Elapsed.TotalSeconds} s");
+
+            return Ok();
+        }
+
+        private async Task<int> CleanupInstancesInternal(List<Instance> instances, List<string> autoDeleteAppIds, string caller = "NightlyCleanup")
```
SandGrainOne, line 179: Remove default nd update call points to set correct name.

src/Storage/Migration/v0.00/03-setup-functions.sql
```diff
@@ -1,4 +1,151 @@
...
+    LANGUAGE 'plpgsql'
+    
+AS $BODY$
+BEGIN
+    IF _sort_ascending IS NULL THEN
+        _sort_ascending := false;
+    END IF;
+
+    RETURN QUERY
+    WITH filteredInstances AS
```
SandGrainOne, line 57: Clean up differences in indentation. Some lines have spaces while other have tabs.

HenningNormann: Done in PR286

src/Storage/Controllers/CleanupController.cs
HenningNormann: Will be removed in PR286

PR #305, opened 2024-01-09
Fixes after TT02 rollout

HenningNormann:
- Added validation of date query parameter
- Updated readdeletedinstances to reflect existing functionality. The db changes will be synced to all environments manually
- Removed adaptive sampling
- Added unit test for appid
- Added retry loop in dataelement to address concurrent access.

### DIFF
--- a/src/Storage/Repository/PgDataRepository.cs
+++ b/src/Storage/Repository/PgDataRepository.cs
@@ -188,7 +188,7 @@ namespace Altinn.Platform.Storage.Repository
             {
                 try
                 {
-                    return await UpdateInternal(instanceGuid, dataElementId, propertylist);
+                    return await UpdateInternal(dataElementId, propertylist);
                 }
                 catch (PostgresException e)
                 {
@@ -207,7 +207,7 @@ namespace Altinn.Platform.Storage.Repository
             throw exceptionToRethrow;
         }
 
-        private async Task<DataElement> UpdateInternal(Guid instanceGuid, Guid dataElementId, Dictionary<string, object> propertylist)
+        private async Task<DataElement> UpdateInternal(Guid dataElementId, Dictionary<string, object> propertylist)
         {
             if (propertylist.Count > 10)
             {
--- a/src/Storage/Repository/PgDataRepository.cs
+++ b/src/Storage/Repository/PgDataRepository.cs
@@ -181,6 +181,34 @@ namespace Altinn.Platform.Storage.Repository
         /// <inheritdoc/>
         public async Task<DataElement> Update(Guid instanceGuid, Guid dataElementId, Dictionary<string, object> propertylist)
         {
+            // Retry loop below because Postgres doesn't lock the element between the read and the following update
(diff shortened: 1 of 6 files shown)


src/Storage/Repository/PgInstanceRepository.cs
```diff
@@ -506,10 +506,17 @@ private static void AddDateParam(string dateParam, StringValues queryValues, Dic
         {
             foreach (string value in queryValues)
             {
-                string @operator = value.Split(':')[0];
-                string dateValue = value[(@operator.Length + 1)..];
-                string postgresParamName = GetPgParamName($"{dateParam}_{@operator}");
-                postgresParams.Add(postgresParamName, valueAsString ? dateValue : DateTimeHelper.ParseAndConvertToUniversalTime(dateValue));
+                try
```
acn-sbuad, line 509: I would suggest we handle this validation at an earlier stage than within the repository class

HenningNormann: Maybe both? If not, the controller should send the data with more structure, so that the repo class wouldn't need to parse the parameters.

src/Storage/Repository/PgInstanceRepository.cs
```diff
@@ -506,10 +506,17 @@ private static void AddDateParam(string dateParam, StringValues queryValues, Dic
...
+                {
+                    string @operator = value.Split(':')[0];
+                    string dateValue = value[(@operator.Length + 1)..];
+                    string postgresParamName = GetPgParamName($"{dateParam}_{@operator}");
+                    postgresParams.Add(postgresParamName, valueAsString ? dateValue : DateTimeHelper.ParseAndConvertToUniversalTime(dateValue));
+                }
+                catch (Exception)
```
SandGrainOne, line 516:
```suggestion
                catch
```

HenningNormann: Ok

src/Storage/Repository/PgInstanceRepository.cs
HenningNormann: I suggest that we don't change existing functionality until the new code has run without problems in production for a couple of weeks.

src/Storage/Repository/PgDataRepository.cs
```diff
@@ -180,6 +180,34 @@ await using (NpgsqlDataReader reader = await pgcom.ExecuteReaderAsync())
...
+                    {
+                        throw;
+                    }
+                }
+            }
+
+            throw exceptionToRethrow;
+        }
+
+        private async Task<DataElement> UpdateInternal(Guid instanceGuid, Guid dataElementId, Dictionary<string, object> propertylist)
```
SandGrainOne, line 210: The `instanceGuid` parameter is unnecessary.

HenningNormann: Done

SandGrainOne (approved):

Issue #284, opened 2023-12-04
Move all automated tests to altinn-storage repository


acn-sbuad:
### Description

Clean up old tests in altinn-studio repository. 

Check what we're testing through use-case testing
Complete confirm
Delete instance, data element
POST, PUT, etc

### Tasks
- [x] Set up maskinporten client in test
- [x] Set up maskinporten client in prod
- [x] Move the test code from altinn-studio to the altinn-storage repository
- [x] Document test coverage and expand as needed.
- [x] Run the tests with github actions instead of Azure DevOps pipelines.
- [x] Set up workflow for at
- [x] Set up workflow for tt
- [x] Set up workflow for prod
- [x] investigate setting up trigger for workflow from devops
- [x] Remove code from altinn-studio repo 

### Acceptance Criterias

- [x] Workflow file available in storage repository 
- [x] Use case tests are triggered every 15 mins from GitHub 
- [x] Code is removed from Studio repo

acn-sbuad:
PR in devops: https://github.com/Altinn/altinn-studio/pull/12072
Also a pending pr in altinn-audit repo

PR #307, opened 2024-01-12
Moved k6 test suite from altinn/altinn-studio repo #GCPActive

acn-sbuad:
## Description

- Moved relevant k6 tests from Altinn studio
- Set up use case workflows to include new tests 
- Set up a regression test workflow for each environment
- Added some scripts we don't want to loose to the archive folder

- All workflows run k6 directly on the agent, not through docker-compose

## Related Issue(s)
- #284

## Verification


tba76 (approved):
🍺 Great work 🍺

PR #312, opened 2024-01-17
Fixes after prod rollout

HenningNormann:
- Removed sampling for exceptions and cleanup
- Added index for deleting data elements (already implemented in prod, tt02, yt01 and at21-24
- Initialized instance to null when creating instance to avoid trying to delete a not existing instance after failure


src/Storage/Migration/v0.00/01-setup-tables.sql
```diff
@@ -68,6 +68,8 @@ TABLESPACE pg_default;
 
 CREATE INDEX IF NOT EXISTS dataelements_instanceinternalId ON storage.dataelements(instanceInternalId);
 CREATE INDEX IF NOT EXISTS dataelements_instanceguid ON storage.dataelements(instanceGuid);
+CREATE INDEX IF NOT EXISTS dataelements_isharddeleted ON storage.dataelements(id)
```
acn-sbuad, line 71: Has this been run manually in all environments?

HenningNormann: Yes, ref comment in the description.

acn-sbuad (approved):
🎈

PR #319, opened 2024-01-19
Remove cosmos from storage

HenningNormann:


src/Storage/Controllers/InstancesController.cs
```diff
@@ -339,7 +339,7 @@
             }
             catch (Exception ex)
             {
-                _logger.LogError(ex, "Something went wrong during GetDecisionForRequest for application id: {appId} AppInfo: {appInfo}", appId, appInfo?.ToString());
+                _logger.LogError(ex, "Something went wrong during GetDecisionForRequest for application id: {appId} AppInfo: {appInfo}", appId, appInfo.ToString());
```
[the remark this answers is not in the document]
SandGrainOne: This is an issue we can look into I think. The input value can be checked and cleaned. We do something similar to file names.

src/Storage/Controllers/CleanupController.cs
```diff
@@ -18,45 +18,26 @@ namespace Altinn.Platform.Storage.Controllers
...
-            ILogger<CleanupController> logger,
-            IOptions<GeneralSettings> generalSettings)
-        {
-            _instanceRepository = instanceRepository;
-            _applicationRepository = applicationRepository;
-            _blobRepository = blobRepository;
-            _dataRepository = dataRepository;
-            _instanceEventRepository = instanceEventRepository;
-            _logger = logger;
-            _usePostgreSQL = generalSettings.Value.UsePostgreSQL;
```
SandGrainOne, line 58: There are unused usings after the cleanup. No need for Configuration for example.

src/Storage/Migration/v0.00/03-setup-functions.sql
```diff
@@ -284,18 +284,6 @@ RETURN QUERY
 END;
 $BODY$;
 
-CREATE OR REPLACE FUNCTION storage.readallformultipledataelement(_instanceGuid UUID[])
```
SandGrainOne, line 287: Do we need a DROP statement in our scripts?

test/UnitTest/Mocks/Repository/InstanceRepositoryMock.cs
```diff
@@ -154,16 +154,16 @@ public Task<(Instance Instance, long InternalId)> GetOne(int instanceOwnerPartyI
             return Task.FromResult<(Instance, long)>((null, 0));
         }
 
-        public Task<Instance> Update(Instance instance)
+        public Task<Instance> Update(Instance item)
```
SandGrainOne, line 157: This change seems counter intuitive.

HenningNormann: It's a complaint from the analyzers because the variable name is "item" in the interface. I can change the interface in the next iteration.

src/Storage/Migration/v0.00/03-setup-functions.sql
HenningNormann: Then we need a new migration folder and script also. I planned to do it manually.

src/Storage/Controllers/CleanupController.cs
HenningNormann: Ok

src/Storage/Migration/v0.00/03-setup-functions.sql
SandGrainOne: okay..

src/Storage/Repository/PgInstanceRepository.cs
```diff
@@ -295,7 +295,7 @@ await using (NpgsqlDataReader reader = await pgcom.ExecuteReaderAsync())
         }
 
         /// <inheritdoc/>
-        public async Task<(Instance Instance, long InternalId)> GetOne(int instanceOwnerPartyId, Guid instanceGuid, bool includeElements = true)
+        public async Task<(Instance Instance, long InternalId)> GetOne(Guid instanceGuid, bool includeElements = true)
```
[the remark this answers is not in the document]
HenningNormann: I think we should change localtest to use the normal postgres implementation rather than a local mock with flat files. Until we do that we can require that the guid is unique across partyid and keep the existing local hierarchy and search in the subdirectories.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -800,5 +802,19 @@ private string BuildRequestLink(string continuationToken = null, Dictionary<stri
                 return selfUrl;
             }
         }
+
+        /// <summary>
+        /// Validates app id
+        /// </summary>
```
SandGrainOne, line 808: No need for comments on private methods. Especially those that don't add anything.

Or change the comment to for example:
Ensures that the app value contains a '/' and that there are no newline characters.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -476,7 +478,7 @@ public async Task<ActionResult<Instance>> Delete(int instanceOwnerPartyId, Guid
             [FromRoute] int instanceOwnerPartyId,
             [FromRoute] Guid instanceGuid)
         {
-            (Instance instance, _) = await _instanceRepository.GetOne(instanceOwnerPartyId, instanceGuid);
+            (Instance instance, _) = await _instanceRepository.GetOne(instanceGuid, true);
```
SandGrainOne, line 481: Maybe add a reason for why includeDataElements are true:
`// Fetching data elements is needed because the action returns the entire instance`

src/Storage/Repository/PgInstanceRepository.cs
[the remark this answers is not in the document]
SandGrainOne: Okay, I gues there exists alternatives still.

SandGrainOne, line 298: Default values can be useful, but completely unnecessary in this case.

SandGrainOne: I get compile errors if I also remove the default value from the interface. I don't like default values in general. It's better to update call points. There aren't that many of them.

src/Storage/Repository/IInstanceRepository.cs
```diff
@@ -23,11 +23,10 @@ public interface IInstanceRepository
         /// <summary>
         /// Get an instance for a given instance id
         /// </summary>
-        /// <param name="instanceOwnerPartyId">the instance owner id</param>
         /// <param name="instanceGuid">the instance guid</param>
         /// <param name="includeElements">whether to include data elements</param>
         /// <returns>The instance for the given parameters</returns>
-        Task<(Instance Instance, long InternalId)> GetOne(int instanceOwnerPartyId, Guid instanceGuid, bool includeElements = true);
+        Task<(Instance Instance, long InternalId)> GetOne(Guid instanceGuid, bool includeElements = true);
```
SandGrainOne, line 29: Avoid default values in cases where the number of call points are manageable.

SandGrainOne (approved):

src/Storage/Controllers/InstancesController.cs
HenningNormann: The must interesting thing to address related to this is how to maintain the two instance properties lastchanged and readstatus. In many use cases the data elements are fetched only to make sure that lastchanged and readstatus are set the same way as with the Cosmos repo. The lastchanged semantics is not obvious - the value in the json doc in the db is not the value used externally. One solution could be to always update the two properties directly on the instance for every data element write operation. Then the semantics of the includeDataelements parameter on GetOne would be obvious.

src/Storage/Repository/IInstanceRepository.cs
HenningNormann: The must interesting thing to address related to this is how to maintain the two instance properties lastchanged and readstatus. In many use cases the data elements are fetched only to make sure that lastchanged and readstatus are set the same way as with the Cosmos repo. The lastchanged semantics is not obvious - the value in the json doc in the db is not the value used externally. One solution could be to always update the two properties directly on the instance for every data element write operation. Then the semantics of the includeDataelements parameter on GetOne would be obvious.

src/Storage/Repository/PgInstanceRepository.cs
HenningNormann: The must interesting thing to address related to this is how to maintain the two instance properties lastchanged and readstatus. In many use cases the data elements are fetched only to make sure that lastchanged and readstatus are set the same way as with the Cosmos repo. The lastchanged semantics is not obvious - the value in the json doc in the db is not the value used externally. One solution could be to always update the two properties directly on the instance for every data element write operation. Then the semantics of the includeDataelements parameter on GetOne would be obvious.

PR #337, opened 2024-02-18
Update instance properties when create/update/delete data element

HenningNormann:
- Implemented PgInstanceRepository.SetStatuses in the db functions/procs for data element create/update/delete
- Updated dataelement.LastChangedBy in the Delete method in DataController
- Added lastChangedBy and isRead to data element test data
- Changed references to appsettings.json in test artifacts  to include path
- Added readstatus to testdata Instance_1_1


src/Storage/Controllers/DataController.cs
```diff
@@ -130,6 +130,7 @@ public async Task<ActionResult<DataElement>> Delete(int instanceOwnerPartyId, Gu
                 return await InitiateDelayedDelete(instance, dataElement);
             }
 
+            dataElement.LastChangedBy = User.GetUserOrOrgId();
```
acn-sbuad, line 133: should we also set it in the case of delayed delete?

test/UnitTest/data/cosmoscollections/dataelements/11f7c994-6681-47a1-9626-fcf6c27308a5.json
```diff
@@ -9,5 +9,7 @@
   "refs": [
   ],
   "created": "2020-05-11T17:09:28.4621953Z",
-  "lastChanged": "2020-05-11T17:09:28.4621953Z"
+  "lastChanged": "2020-05-11T17:09:28.4621953Z",
```
acn-sbuad, line 12: Is it still relevant to update a cosmoscollection of testdata? Do we need to rename the folder?

src/Storage/Controllers/DataController.cs
HenningNormann: Done

acn-sbuad (approved):
🎈

PR #340, opened 2024-02-22
only running hide instancs if all data present

acn-sbuad:
## Description

Only attempt to hide instances if application metadata available to check settings


tba76 (approved):
🍺

PR #341, opened 2024-02-23
Change data element update to patch json and fix timestamp issue

HenningNormann:
- Changed dataelement update from read/update-full-json to updating only changed json properties
- Changed the delete data element functions to store timestamps with Z rather than + to avoid system.text.json deserialization bug (ignoring time zone and generating DateTime of kind local).


acn-sbuad (approved):
🎈

PR #344, opened 2024-02-26
Replaced space with "T" when setting lastchanged to now() in the delete data element db functions

HenningNormann:


acn-sbuad (approved):
🎈

PR #345, opened 2024-02-27
Reduced fetch of data elements after change to maintain readstatus/lastchanged* when changing data elements

HenningNormann:
- Added includeDataelements as parameter to GetInstancesFromQuery. Should be false from sbl message box search.
- Made parameter includeElements in IInstanceRepository.GetOne mandatory
- Changed GetMessageBoxInstance to not get elements
- Avoid updating read status if status is unchanged
- Removed SetStatues (previous PR ensure statuses maintained for data element create/write/delete)
- Reduced number of warnings and info messages
- Added unit test for continuation token
- Added more logging functionality to TelemetryTracker
- Added logging to GetInstancesFromQuery
- Added null-check in GetSBLStatusForCurrentTask

### DIFF
--- a/src/Storage/Helpers/InstanceHelper.cs
+++ b/src/Storage/Helpers/InstanceHelper.cs
@@ -98,6 +98,7 @@ namespace Altinn.Platform.Storage.Helpers
         {
             if (instance.Process != null)
             {
+                string taskType;
                 if (instance.Process.Ended != null && instance.Status?.Archived == null)
                 {
                     return "Submit";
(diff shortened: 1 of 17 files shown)


src/Storage/Helpers/InstanceHelper.cs
```diff
@@ -106,22 +106,23 @@ public static string GetSBLStatusForCurrentTask(Instance instance)
...
-                    return "Feedback";
-                }
-                else if (instance.Process.CurrentTask.AltinnTaskType.Equals("signing", StringComparison.OrdinalIgnoreCase))
-                {
-                    return "Signing";
-                }
-                else
-                {
-                    return "FormFilling";
+                    if (instance.Process.CurrentTask.AltinnTaskType.Equals("confirmation", StringComparison.OrdinalIgnoreCase))
```
acn-sbuad, line 111: could add a local variable string TaskType = instance.Process.CurrentTask.AltinnTaskType; 
to make reading the following if checks a bit easier.

src/Storage/Repository/PgInstanceRepository.cs
```diff
@@ -20,11 +20,12 @@ namespace Altinn.Platform.Storage.Repository
     /// </summary>
     public class PgInstanceRepository: IInstanceRepository
     {
+        private const string _readSqlFilteredInitial = "select * from storage.readinstancefromquery_v2 (";
         private readonly string _deleteSql = "select * from storage.deleteinstance ($1)";
         private readonly string _insertSql = "call storage.insertinstance ($1, $2, $3, $4, $5, $6, $7, $8)";
         private readonly string _upsertSql = "call storage.upsertinstance ($1, $2, $3, $4, $5, $6, $7, $8)";
         private readonly string _readSql = "select * from storage.readinstance ($1)";
-        private readonly string _readSqlFiltered = "select * from storage.readinstancefromquery (";
+        private readonly string _readSqlFiltered = _readSqlFilteredInitial;
```
acn-sbuad, line 28: Could the sql be defined directly on this lin instead?

acn-sbuad (approved):
🎈

src/Storage/Repository/PgInstanceRepository.cs
HenningNormann: This is to avoid having and maintaining a very long string literal in the code.

src/Storage/Helpers/InstanceHelper.cs
HenningNormann: Done

PR #349, opened 2024-03-04
Set DependencyTelemetry sampling % to 100 and fix update of instance lastChanged

HenningNormann:
- Converted lastChanged so that the precision is always 6 digits
- Corrected bug in PgDataRepository: Instance object wasn't populated
- Set DependencyTelemetry sampling % to 100
- Changed storage.insertdataelement from procedure to function to return updated dataelement.
- Changed storage.updatedataelement_v2 so that lastChanged json property gets av value with 6 digits. Also updated PgDataRepository to reflect the updated element
- Added functionality to unit tests for data element create and update to check the updated lastchanged values in instance and dataelement

### DIFF
--- a/src/Storage/Configuration/CustomTelemetryInitializer.cs
+++ b/src/Storage/Configuration/CustomTelemetryInitializer.cs
@@ -20,13 +20,13 @@ namespace Altinn.Platform.Telemetry
                 telemetry.Context.Cloud.RoleName = "platform-storage";
             }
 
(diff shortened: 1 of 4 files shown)


acn-sbuad (approved):
🎈 

Would some integration tests where we verify the length of the time stamps be useful here?

HenningNormann:
> 🎈
> 
> Would some integration tests where we verify the length of the time stamps be useful here?

HenningNormann:
Time stamp verification is added to the unit tests

PR #352, opened 2024-03-11
Changed updating of instance from read and update in different transactions to atomic update with json merge.

HenningNormann:
Updating of DataValues and ConfirmConfirmations must be treated specially because the postgres json merge operator (||) doesn't support merging of objects/arrays (the right operand overwrites values in the left operand).

Unit tests are added to verify different update combinations.


acn-sbuad (approved):
🎈

PR #355, opened 2024-03-18
Autogenerate db script for deploying db functions and procedures

HenningNormann:
## Description
Suggested short term deploy enhancement for functions/procs.

Rules:
- Each function/proc has a separate file in the Migration/FunctionsAndProcedures folder.
- After build the tool DbTools.exe is automatically run to generate vx.xx/ZZ-functions-and-procedures.sql based on the content in all files with a newer last changed timestamp than the created timestamp of the newest vx.xx folder. ZZ is assumed to be 01 if no other file is found with another number. (Makes it possible to deploy other stuff before procs/funcs.)
- The file name is the base proc/function name without any version postfix. The same filename is kept if the proc/func gets a new version.
- Any drop commands must be coded at the top of the func/proc file or in a separate file in the related v.x.xx folder.
- A new vx.xx folder must be created when a func/proc is created/updated after last deploy. If not the current vx.xx will be used, and the migration will not be executed by yuniql.

Fixed bug for updating substatus (lacked setting update properties for description and label)

## Related Issue(s)
- #367 

## Verification
- [ ] All tests run green

## Documentation


test/UnitTest/Altinn.Platform.Storage.UnitTest.csproj
```diff
@@ -26,153 +26,174 @@
...
-    <Content Remove="data\response_multi_permit.json" />
-    <Content Remove="data\response_permit.json" />
-    <Content Remove="data\roles\user_1\party_1000\roles.json" />
-    <Content Remove="data\roles\user_1\party_1600\roles.json" />
-    <Content Remove="data\roles\user_1\party_50000000\roles.json" />
-    <Content Remove="data\roles\user_1\party_500\roles.json" />
-    <Content Remove="data\roles\user_3\party_1606\roles.json" />
-    <Content Remove="data\roles\user_3\party_1600\roles.json" />
-  </ItemGroup>
+	<ItemGroup>
```
acn-sbuad, line 63: could this section be deleted a a whole? Don't think the files are in the repository any more

acn-sbuad (approved):
🎈 
How do we best communicate the new routine? Should we add a readme to the Migration folder for instance?

Issue #331, opened 2024-02-01
Optimize storage instance properties


HenningNormann:
The storage instance properties lastchanged and readstatus are currently being populated by reading all data elements. These properties should be updated directly on the instance when a relevant data element event occurs.

Could be done in three iterations:

1. Deploy change to code and db procedures to update instance when doing data element create, update and delete
2. Convert existing instances
3. Deploy change:
    1. Remove old code to set properties
    2. Remove old procedures
    3. Remove join to data elements for get instance and get instances
    4. Change GetOne to remove getDataElements parameter
    5. Update unit tests

PR #359, opened 2024-03-22
Changed instance update to reflect various json merge needs

HenningNormann:
Changed instance update to reflect various json merge needs. Cases:
- Process: Insert complete json
- Substatus: Insert complete json
- Status: Merge propertis based on updateProperties in repo api
- CompleteConfirmations: Merge properties
- Datavalues: Merge properties
- PresentationTexts: Merge properties
- Other top level properties (lastchanged and lastchangedby): Merge properties
- "Complications": Process and Substatus might be combined, and the different combinations must be treated separately

## Related Issue(s)
- #331

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Migration/v0.06/01-functions-and-procedures.sql
+++ b/src/Storage/Migration/v0.06/01-functions-and-procedures.sql
@@ -1,0 +1,135 @@
+CREATE OR REPLACE FUNCTION storage.updateinstance_v2(
+        _alternateid UUID,
+        _toplevelsimpleprops JSONB,
+        _datavalues JSONB,
+        _completeconfirmations JSONB,
+        _presentationtexts JSONB,
+        _status JSONB,
+        _substatus JSONB,
+        _process JSONB,
+        _lastchanged TIMESTAMPTZ,
+        _taskid TEXT)
+    RETURNS TABLE (updatedInstance JSONB)
+    LANGUAGE 'plpgsql'	
+AS $BODY$
+BEGIN
+    IF _datavalues IS NOT NULL THEN
+        RETURN QUERY
+            UPDATE storage.instances SET
+                instance = instance || _toplevelsimpleprops ||
+                    jsonb_strip_nulls(
+                        jsonb_set(
+                            '{"DataValues":""}',
+                            '{DataValues}',
+                            CASE WHEN instance -> 'DataValues' IS NOT NULL THEN
+                                instance -> 'DataValues' || _datavalues
+                            ELSE
+                                _datavalues
+                            END
+                        )
+                    ),
+                lastchanged = _lastchanged
+            WHERE _alternateid = alternateid
+            RETURNING instance;
+    ELSIF _presentationtexts IS NOT NULL THEN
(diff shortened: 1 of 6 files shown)


src/Storage/Migration/v0.06/01-functions-and-procedures.sql
```diff
@@ -0,0 +1,135 @@
...
+        _status JSONB,
+        _substatus JSONB,
+        _process JSONB,
+        _lastchanged TIMESTAMPTZ,
+        _taskid TEXT)
+    RETURNS TABLE (updatedInstance JSONB)
+    LANGUAGE 'plpgsql'	
+AS $BODY$
+BEGIN
+    IF _datavalues IS NOT NULL THEN
```
SandGrainOne, line 16: One observation. It's not possible to update datavalues and process (examples) in one call. There probably aren't any need for that exact example, but have we considered all permutations?

SandGrainOne: I'm slightly worried it's possible to use this incorrectly.

test/UnitTest/TestingRepositories/InstanceTests.cs
```diff
@@ -72,7 +74,48 @@ public async Task Instance_Update_Task_Ok()
             int count = await PostgresUtil.RunCountQuery(sql);
             Assert.Equal(1, count);
             Assert.Equal("Task_2", updatedInstance.Process.CurrentTask.ElementId);
-            Assert.NotEqual("Shouldn't be updated", updatedInstance.Process.CurrentTask.Name);
+            Assert.Equal(newInstance.Process.CurrentTask.Name, updatedInstance.Process.CurrentTask.Name);
+            Assert.Equal("After update", updatedInstance.Process.CurrentTask.Name);
```
SandGrainOne, line 78: Could you add some asserts for properties we expect to remain unchanged?

Insert a value for `newInstance.Process.CurrentTask.AltinnTaskType` and check that it's the same in the updated instance.

src/Storage/Migration/v0.06/01-functions-and-procedures.sql
HenningNormann: I have tried to cover all current combinations. Process is supported both stand alone and combined with status. (And lastchanged* is added to all combinations.)

test/UnitTest/TestingRepositories/InstanceTests.cs
HenningNormann: The PR reflects the functionality in the current code. When updating the Process object all properties are set from user input and/or controller logic. No info from the existing instance will survice.

SandGrainOne: The bug we're trying to fix with this change was an issue with IsHardDeleted under Status disapearing after an update to IsArchived or vise versa. I want a test for this case.

HenningNormann: This is already covered on line 114

PR #360, opened 2024-03-25
#3 Adding the option of searching for instances based on person or org number, or other in storage

codemascot:
Adding the option of searching for instances based on person or org number, or other in storage

## Description
Adding the option of searching for instances based on person or org number, or other in storage

## Related Issue(s)
- #3

## Verification
- [x] All tests run green

## Documentation


src/Storage/Controllers/InstancesController.cs
```diff
@@ -155,7 +161,22 @@ public class InstancesController : ControllerBase
...
+
+                    if (string.IsNullOrEmpty(instanceOwnerIdType) || string.IsNullOrEmpty(instanceOwnerIdValue))
+                    {
+                        return BadRequest("Invalid InstanceOwnerIdentifier.");
+                    }
+
+                    string orgNo = instanceOwnerIdType == "organization" ? instanceOwnerIdValue : string.Empty;
+                    string person = instanceOwnerIdType == "person" ? instanceOwnerIdValue : string.Empty;
+
+                    instanceOwnerPartyId = await _registerService.PartyLookup(orgNo, person);
```
SandGrainOne, line 179: This value is actually never used. See the warning from Github below. The parameter values are being read twice. We do some basic validation here in the Controller, but the final values are taken from the original request. 

See line 193: `Dictionary<string, StringValues> queryParams = QueryHelpers.ParseQuery(Request.QueryString.Value);`

We might need to expand the _instanceRepository.GetInstancesFromQuery method with a new parameter for the partyId and update the method to ignore the queryParams value for party id.

codemascot: Alternative solution: https://github.com/Altinn/altinn-storage/pull/360/commits/fd47008e8351cc5e40cc82758abae33796ec5884

test/UnitTest/Extensions/HttpClientExtensionTest.cs
```diff
@@ -0,0 +1,90 @@
...
+        public async Task PostAsync_ShouldAddAuthorizationHeaderAndReturnHttpResponseMessage()
+        {
+            // Arrange
+            HttpResponseMessage httpResponseMessage = new HttpResponseMessage
+            {
+                StatusCode = HttpStatusCode.OK,
+                Content = new StringContent("dummyData", Encoding.UTF8, "application/json")
+            };
+
+            Mock<HttpMessageHandler> handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
```
SandGrainOne, line 28: Instead of mocking a HttpMessageHandler like this, we've started implementing a [DelegatingHandlerStub](https://github.com/Altinn/altinn-notifications/blob/main/test/Altinn.Notifications.Tests/DelegatingHandlerStub.cs) class. Take a look at how we're using it in the notifications solution.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -155,7 +162,22 @@ public class InstancesController : ControllerBase
             {
                 if (instanceOwnerPartyId == null)
                 {
-                    return BadRequest("InstanceOwnerPartyId must be defined.");
+                    if (string.IsNullOrEmpty(instanceOwnerIdentifier))
+                    {
+                        return BadRequest("Either InstanceOwnerPartyId or InstanceOwnerIdentifier must need to be defined.");
```
SandGrainOne, line 167:
```suggestion
                        return BadRequest("Either InstanceOwnerPartyId or InstanceOwnerIdentifier need to be defined.");
```

src/Storage/Services/RegisterService.cs
```diff
@@ -0,0 +1,115 @@
...
+            HttpResponseMessage response = await _client.PostAsync(bearerToken, endpointUrl, content, accessToken);
+            if (response.StatusCode == HttpStatusCode.OK)
+            {
+                Party party = await response.Content.ReadFromJsonAsync<Party>(_serializerOptions);
+                return party.PartyId;
+            }
+            else
+            {
+                string reason = await response.Content.ReadAsStringAsync();
+                _logger.LogError("// RegisterService // PartyLookup // Failed to lookup party in platform register. Response {response}. \n Reason {reason}.", response, reason);
```
SandGrainOne, line 109: I don't want us to serialize the full response into an error message like this. Limit it to the response code. The body of the respons (reason) can also be pretty large, but let's keep it for now.

SandGrainOne: What do you think of the warning below?

codemascot: Regarding the `PascalCase` warning, of course this need to be fixed, but it actually is like this in `altinn-events`, hereby it is kept this way to make the coding standard comply with other Altinn projects.

test/UnitTest/Extensions/HttpClientExtensionTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        private readonly HttpClient _httpClient;
+        private HttpRequestMessage _httpRequest;
+
+        public HttpClientExtensionTests()
+        {
+            var httpMessageHandler = new DelegatingHandlerStub(async (request, token) =>
+            {
+                _httpRequest = request;
+                await Task.Yield();
+                return new HttpResponseMessage(HttpStatusCode.OK);
```
SandGrainOne, line 23: This feels safer. It's something we've already used quite a bit:
```suggestion
                return await Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));
```

SandGrainOne (approved):

src/Storage/Configuration/PlatformSettings.cs
```diff
@@ -0,0 +1,31 @@
+using System.Runtime.Serialization;
+
+namespace Altinn.Platform.Storage.Configuration
+{
+    /// <summary>
+    /// Represents a set of configuration options when communicating with the platform API.
+    /// Instances of this class is initialised with values from app settings. Some values can be overridden by environment variables.
+    /// </summary>
+    public class PlatformSettings
```
acn-dgopa, line 9: @acn-sbuad @codemascot I suspect that the problem lies here. I'm not sure how the platformsettings was handled before. Now i see that this file does not have ApiAuthorizationEndpoint and therefore platformsettings does not read the config value ApiAuthorizationEndpoint. So when PEP tries to fetch the ApiAuthorizationEndpoint, its null. Therefore the uri format exception.

src/Storage/Program.cs
```diff
@@ -189,12 +190,13 @@ void ConfigureServices(IServiceCollection services, IConfiguration config)
...
 
     services.AddHttpClient<AuthorizationApiClient>();
+    services.AddHttpClient<IRegisterService, RegisterService>();
 
     services.Configure<AzureStorageConfiguration>(config.GetSection("AzureStorageConfiguration"));
     services.Configure<GeneralSettings>(config.GetSection("GeneralSettings"));
     services.Configure<KeyVaultSettings>(config.GetSection("kvSetting"));
     services.Configure<PepSettings>(config.GetSection("PepSettings"));
-    services.Configure<PlatformSettings>(config.GetSection("PlatformSettings"));
+    services.Configure<Altinn.Platform.Storage.Configuration.PlatformSettings>(config.GetSection("PlatformSettings"));
```
SandGrainOne, line 199: @codemascot Here we should have added a new line instead of changing the one we had. We can also rename the new type from PlatformSettings to RegisterClientSettings to avoid the ambiguity.

src/Storage/Configuration/PlatformSettings.cs
codemascot: Thanks for pointing that out! 👍 Here is the fix commit: https://github.com/Altinn/altinn-storage/pull/376/commits/25aba89a59880ff8632d11269a6d696d6cb0be04 🙂

src/Storage/Program.cs
codemascot: Correct! 👍 Here is the implementation commit: https://github.com/Altinn/altinn-storage/pull/376/commits/25aba89a59880ff8632d11269a6d696d6cb0be04 🙂

PR #381, opened 2024-04-18
#3 k6 tests added for the new feature

codemascot:
#3 k6 tests added for the new feature

## Description
#3 k6 tests added for the new feature

## Related Issue(s)
- #3 

## Verification
- [x] All tests run green

### DIFF
--- a/.github/workflows/regression-test-ATX.yml
+++ b/.github/workflows/regression-test-ATX.yml
@@ -40,7 +40,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e tokenGeneratorUserName=${{ secrets.TOKENGENERATOR_USERNAME }} -e tokenGeneratorUserPwd=${{ secrets.TOKENGENERATOR_USERPASSWORD }} -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e useTestTokenGenerator=true
+        flags: -e runFullTestSet=true -e tokenGeneratorUserName=${{ secrets.TOKENGENERATOR_USERNAME }} -e tokenGeneratorUserPwd=${{ secrets.TOKENGENERATOR_USERPASSWORD }} -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e useTestTokenGenerator=true
     - name: Run regression tests for endpoint 'instances/sbl'
       if: always()
       uses: grafana/k6-action@v0.3.1
--- a/.github/workflows/regression-test-ATX.yml
+++ b/.github/workflows/regression-test-ATX.yml
@@ -40,7 +40,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e tokenGeneratorUserName=${{ secrets.TOKENGENERATOR_USERNAME }} -e tokenGeneratorUserPwd=${{ secrets.TOKENGENERATOR_USERPASSWORD }} -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e useTestTokenGenerator=true
+        flags: -e runFullTestSet=true -e tokenGeneratorUserName=${{ secrets.TOKENGENERATOR_USERNAME }} -e tokenGeneratorUserPwd=${{ secrets.TOKENGENERATOR_USERPASSWORD }} -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e orgPartyId=${{ secrets.ORG_PARTY_ID }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e useTestTokenGenerator=true
     - name: Run regression tests for endpoint 'instances/sbl'
       if: always()
       uses: grafana/k6-action@v0.3.1
--- a/.github/workflows/regression-test-PROD.yml
+++ b/.github/workflows/regression-test-PROD.yml
@@ -21,7 +21,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
+        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e orgPartyId=${{ secrets.ORG_PARTY_ID }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
 
     - name: Run regression tests for endpoint 'instances/sbl'
       if: always()
--- a/.github/workflows/regression-test-PROD.yml
+++ b/.github/workflows/regression-test-PROD.yml
@@ -21,7 +21,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
+        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
 
     - name: Run regression tests for endpoint 'instances/sbl'
       if: always()
--- a/.github/workflows/regression-test-PROD.yml
+++ b/.github/workflows/regression-test-PROD.yml
@@ -21,7 +21,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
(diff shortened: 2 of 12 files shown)


test/k6/src/tests/instances.js
```diff
@@ -291,6 +297,50 @@ function TC11_HardDeleteInstance(data) {
...
+  const instanceIdSplit = firstInstance["id"].split("/");
+
+  success = check(res, {
+    "TC12_GetInstances_ByPersonNumber: Get instance for party. InstanceId has expected format":
+      instanceIdSplit.length === 2,
+  });
+  addErrorCount(success);
+}
+
+//TC13 - Get all instances for party looked up with a person number
```
acn-sbuad, line 322:
```suggestion
//TC13 - Get all instances for party looked up with an organisation number
```

test/k6/src/tests/instances.js
```diff
@@ -291,6 +297,50 @@ function TC11_HardDeleteInstance(data) {
...
+    "TC13_GetInstances_ByOrgNumber: Get instance for party. Status is 200": (
+      r
+    ) => r.status === 200,
+  });
+  addErrorCount(success);
+
+  const firstInstance = JSON.parse(res.body).instances[0];
+  const instanceIdSplit = firstInstance["id"].split("/");
+
+  success = check(res, {
```
acn-sbuad, line 337: A more interesting check here might be to check instanceOwner.organisationNumber to see if this matches the org number used as the instance owner identifier.

acn-sbuad:
Updates to workflows to pass new environment variables is missing.

test/k6/src/tests/instances.js
```diff
@@ -5,6 +5,7 @@
     -e env=*** `
     -e userId=*** `
     -e partyId=*** `
+    -e orgNumber=*** `
```
acn-sbuad, line 8: I would add org number below existing user params. As to not create confusion of who the user and partyIds are linked to.

.github/workflows/regression-test-ATX.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e tokenGeneratorUserName=${{ secrets.TOKENGENERATOR_USERNAME }} -e tokenGeneratorUserPwd=${{ secrets.TOKENGENERATOR_USERPASSWORD }} -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e useTestTokenGenerator=true
+        flags: -e runFullTestSet=true -e tokenGeneratorUserName=${{ secrets.TOKENGENERATOR_USERNAME }} -e tokenGeneratorUserPwd=${{ secrets.TOKENGENERATOR_USERPASSWORD }} -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e userId=${{ secrets.USER_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e useTestTokenGenerator=true
```
acn-sbuad, line 43: Use case tests workkflow will also need updating with the same parameters :)

acn-sbuad: TT and prod are still missing here. Reopening

test/k6/src/api/instances.js
```diff
@@ -42,6 +42,14 @@ export function getInstances(token, filters) {
   return http.get(endpoint, params);
 }
 
+//Api call to Storage:Instances to get an instance by person number and return response
```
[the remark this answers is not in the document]
codemascot: Is there a specific rationale for this approach? 🤔 My reasoning for creating a separate function is to enhance maintainability and adaptability for new test cases based on the feature/purpose, without hampering the flow of already existing ones. And on the other hand, isn't adding new parameters makes the code more complex? Considering these scenarios, blending this to existing function do we achieve any particular benefit? 🤔

.github/workflows/regression-test-PROD.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
+        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.REGRESSION_USER_NAME}} -e userPassword=${{ secrets.REGRESSION_USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
```
acn-sbuad, line 24: why is partyId required here?

.github/workflows/regression-test-TT02.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.USE_CASE_USER_NAME}} -e userPassword=${{ secrets.USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
+        flags: -e runFullTestSet=true -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e partyId=${{ secrets.PARTY_ID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.USE_CASE_USER_NAME}} -e userPassword=${{ secrets.USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
```
acn-sbuad, line 24: Same goes here. Why is partyId required?

test/k6/src/tests/instances.js
```diff
@@ -291,6 +295,53 @@ function TC11_HardDeleteInstance(data) {
   addErrorCount(success);
 }
 
+//TC12 - Get all instances for party looked up with a person number
+function TC12_GetInstances_ByPersonNumber(data) {
+  var instanceOwnerIdentifier = "Person:" + data.personNumber;
```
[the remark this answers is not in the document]
acn-sbuad: Unresolving. Adding a condition only means if we don't get any instances returned we don't check for the existence of instances either.. In it's current state this test does in no way verify that the lookup functionality is working as expected. 

If no instance is returned in the list the test should fail.

.github/workflows/regression-test-PROD.yml
codemascot: Yea! Correct! We don't need party ID when using basic auth! 👍

acn-sbuad: so the property should be removed. Unresolving

test/k6/src/tests/instances.js
```diff
@@ -276,15 +280,83 @@ function TC10_SoftDeleteInstance(data) {
...
+
+    success = check(res, {
+      "TC11_GetInstances_ByPersonNumber: Get instance for party. InstanceId has expected format":
+        instanceIdSplit.length === 2,
+    });
+    addErrorCount(success);
+  }
+}
+
+//TC13 - Get all instances for party looked up with an organisation number
```
acn-sbuad, line 314: mismatch tc12 and t13

test/k6/src/tests/instances.js
```diff
@@ -276,15 +280,83 @@ function TC10_SoftDeleteInstance(data) {
...
+function TC12_GetInstances_ByOrgNumber(data) {
+  var instanceOwnerIdentifier = "Organisation:" + data.orgNumber;
```
acn-sbuad, line 316: are we ensuring before this tests runs that there actually exists any instances for the given orgnumber?

test/k6/src/tests/instances.js
```diff
@@ -276,15 +280,83 @@ function TC10_SoftDeleteInstance(data) {
...
+  var success = check(res, {
+    "TC12_GetInstances_ByOrgNumber: Get instance for party. Status is 200": (
+      r
+    ) => r.status === 200,
+  });
+  addErrorCount(success);
+
+  const firstInstance = JSON.parse(res.body).instances[0];
+  if (firstInstance) { // Making sure there actually exists instances for the instanceOwner
+    const orgnasationNumber = firstInstance.instanceOwner.organisationNumber;
```
acn-sbuad, line 334: typo `orgnasationNumber`

codemascot: This is resolved!

.github/workflows/regression-test-PROD.yml
codemascot: My bad! Double checked this time! `-e partyId=${{ secrets.PARTY_ID }}` removed from all the 4 files. 👍

.github/workflows/regression-test-TT02.yml
codemascot: https://github.com/Altinn/altinn-storage/pull/381#discussion_r1611101856

test/k6/src/tests/instances.js
```diff
@@ -276,15 +280,83 @@ function TC10_SoftDeleteInstance(data) {
   addErrorCount(success);
 }
 
-function TC11_HardDeleteInstance(data) {
+//TC12 - Get all instances for party looked up with a person number
```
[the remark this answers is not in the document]
codemascot: Checked! Now this has been fixed! Marking this resolved! 👍

test/k6/src/tests/instances.js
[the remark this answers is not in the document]
codemascot: Checked! 👍 Updated with a mechanism to test the existence of the instances.

test/k6/src/tests/instances.js
```diff
@@ -73,12 +77,17 @@ export function setup() {
 
 // TC01 - POST instance
 function TC01_PostInstance(data) {
+  const options = {
+    "personNumber": data.personNumber
+  };
+
   var res = instancesApi.postInstance(
-    data.userToken,
+    data.orgToken,
```
acn-sbuad, line 85: no need to change this to org token. The user is authorized to instantiate on behalf of themselves.

test/k6/src/tests/instances.js
```diff
@@ -73,12 +77,17 @@ export function setup() {
 
 // TC01 - POST instance
 function TC01_PostInstance(data) {
+  const options = {
```
acn-sbuad, line 80: Would reconsider including the options here. Look into how partyId and appId is set for the instance before sending to storage. Could personNumber be handled the same way? Or should we rather include all the properties in options? Using both approaches seem a bit messy.

test/k6/src/tests/instances.js
```diff
@@ -276,15 +285,134 @@ function TC10_SoftDeleteInstance(data) {
...
+  addErrorCount(instanciationResultSuccess);
+
+  if (dataBody.count > 0) { // If instance(s) exists
+    const firstInstance = dataBody.instances[0];
+    const personNumber = firstInstance.instanceOwner.personNumber;
+    const instanceIdSplit = firstInstance["id"].split("/");
+
+    const success = check(res, {
+      "TC11_GetInstances_ByPersonNumber: Get instance for party. InstanceId has expected format":
+        instanceIdSplit.length === 2,
```
acn-sbuad, line 317: once we get to this test we are already confident the format is as expected as this is covered by an earlier TC.  Can limit the number of checks to what is relevant for the feature being tested.

test/k6/src/tests/instances.js
codemascot: Correct, as this is for the person number test.

test/k6/src/tests/instances.js
codemascot: I would go for putting all the properties in options as it seems easier to handle validating specific parameters. Plus, option represents both `personNumber` and `orgNumber` which seemed a bit easy to handle than passing two parameters when calling this function. Hereby, my money is on passing the parameters as an object rather separate separate parameters. 🤔

test/k6/src/tests/instances.js
```diff
@@ -276,15 +286,130 @@ function TC10_SoftDeleteInstance(data) {
...
+    ) => r.status === 200,
+    "TC11_GetInstances_ByPersonNumber: Instances exists for party.": 
+    dataBody.count > 0,
+  });
+  addErrorCount(instanciationResultSuccess);
+
+  if (dataBody.count > 0) { // If instance(s) exists
+    const firstInstance = dataBody.instances[0];
+    const personNumber = firstInstance.instanceOwner.personNumber;
+    const instanceIdSplit = firstInstance["id"].split("/");
```
acn-sbuad, line 314: constant is not used anywhere

test/k6/src/tests/instances.js
```diff
@@ -276,15 +286,129 @@ function TC10_SoftDeleteInstance(data) {
...
+    const successHardDelete = check(hardDeletionResult, {
+      "TC12_GetInstances_ByOrgNumber: Hard delete instance for organisation. Status is 200": (r) =>
+        r.status === 200,
+    });
+  
+    addErrorCount(successHardDelete);
+  }
+}
+
+// Here we are deleting instances,
```
acn-sbuad, line 399: no need for this comment

.github/workflows/use-case-TT02.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       uses: grafana/k6-action@v0.3.1
       with:
         filename: test/k6/src/tests/instances.js
-        flags: -e env=${{ vars.ENV }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.USE_CASE_USER_NAME}} -e userPassword=${{ secrets.USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
+        flags: -e env=${{ vars.ENV }} -e pid=${{ secrets.PERSON_NUMBER }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e orgPartyId=${{ secrets.ORG_PARTY_ID }} -e org=${{ vars.ORG }} -e app=${{ vars.APP }} -e userName=${{ secrets.USE_CASE_USER_NAME}} -e userPassword=${{ secrets.USER_PASSWORD}} -e apimSubsKey=${{ secrets.APIM_SUBSKEY }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e encodedJwk=${{ secrets.MP_ENCODEDJWK }}
```
acn-sbuad, line 25: I don't think we need to add the org tests to the use case tests. Would suffice to have them in the regression test and only include one of the new tests in the use case tests

test/k6/src/tests/instances.js
```diff
@@ -9,6 +9,8 @@
     -e username=*** `
     -e userpwd=*** `
     -e org=ttd `
+    -e orgNumber=*** `
```
acn-sbuad, line 12: let's not split up org, app which are naturally grouped together here. OrgNumber and orgPartyId do not represent the same entity as org. New properties could be placed before org instead?

.github/workflows/use-case-TT02.yml
acn-sbuad: you've removed the passing of environment variables, but you will also need to update the test file so test case isn't run outside of the full test set.

test/k6/src/api/instances.js
```diff
@@ -5,40 +5,52 @@ import { stopIterationOnFail } from "../errorhandler.js";
 
 /**
  * Api call to Storage:Instances to create an app instance and returns response
- * @param {*} token token value to be sent in apiHelper for authentication
- * @param {*} partyId party id of the user to whom instance is to be created
- * @param {*} org short name of ap owner
- * @param {*} app app name
- * @param {JSON} serializedInstance instance json metadata sent in request body
- * @returns {JSON} Json object including response apiHelperss, body, timings
+ * @param {Object} instanciationOptions Object containing the parameters for the API call.
```
acn-sbuad, line 8: spelling error `instanciation` should be `instantiation`

acn-sbuad, line 8: spelling error `instanciation` should be `instantiation`

codemascot: What a silly mistake! I should have been more careful! My bad! Fixed now! 👍 Marking is resolved. 🙂

test/k6/src/tests/instances.js
```diff
@@ -276,15 +286,125 @@ function TC10_SoftDeleteInstance(data) {
...
+    });
+    addErrorCount(hasInstances);
+  
+    if (dataBody.count > 0) { // If instance(s) exists
+      const firstInstance = dataBody.instances[0];
+      const orgNumber = firstInstance.instanceOwner.organisationNumber;
+      const instanceIdSplit = firstInstance["id"].split("/");
+  
+      const successResponse = check(res, {
+        "TC12_GetInstances_ByOrgNumber: Get instance for party. InstanceId has expected format":
```
acn-sbuad, line 375: This check has already been executed in a previous test. No need to include here as well

test/k6/src/archive/rf0002.js
```diff
@@ -51,8 +51,16 @@ export default function () {
   const partyId = data['orgPartyId'];
   instanceId = '';
 
-  //Test to create an instance with storage api and validate the response
-  res = instances.postInstance(runtimeToken, partyId, appOwner, level2App, instanceJson);
+  //Test to create an instance with storage api and validate the response // token, partyId, org, app, serializedInstance
```
acn-sbuad, line 54:
```suggestion
  //Test to create an instance with storage api and validate the response 
```

acn-sbuad (approved):
🎈

PR #385, opened 2024-04-22
#3 Both instance owner identifier header and party id cannot be present at the same time

codemascot:
Both person header and party id cannot be present at the same time

## Description
Both person header and party id cannot be present at the same time

## Related Issue(s)
- #3 

## Verification
- [x] All tests run green


src/Storage/Controllers/InstancesController.cs
```diff
@@ -136,6 +136,11 @@ public class InstancesController : ControllerBase
             string orgClaim = User.GetOrg();
             int? userId = User.GetUserIdAsInt();
 
+            if (instanceOwnerPartyId.HasValue && !string.IsNullOrEmpty(instanceOwnerIdentifier))
+            {
+                return BadRequest("Both InstanceOwnerPartyId and InstanceOwnerIdentifier cannot be present at the same time.");
```
acn-sbuad, line 141: Adding the '.' to get the same notation as query parameter
```suggestion
                return BadRequest("Both InstanceOwner.PartyId and InstanceOwnerIdentifier cannot be present at the same time.");
```

Issue #351, opened 2024-03-11
Signature should add reference to current task


tjololo:
This issue has been investigated and is ready for implementation via the following PRs:
- https://github.com/Altinn/altinn-storage/pull/392
- https://github.com/Altinn/altinn-storage/pull/404
- https://github.com/Altinn/app-lib-dotnet/pull/617

### Description

DataElement can add [references](https://github.com/Altinn/altinn-storage/blob/main/src/Storage.Interface/Models/DataElement.cs#L116). 
When Storage generates a signature and stores it in a DataElement it should add a reference with the values: 
Value: `Intstance.Process.CurrentTask.ElementId` 
Relation: `RelationType.GeneratedFrom`
ValueType: `ReferenceType.Task`

This will ensure that if a user navigates back/forward old signature objects will be deleted when a signature task starts

PR #392, opened 2024-04-25
Add task reference to signatures [part 1/2]

danielskovli:
## Description
`DataElements` created during a signature step should be annotated with the `task` they belong to (were generated from).

Establishing the signature <-> task link enables cleanup and introspection where required, eg. to prevent a signature task from generating duplicate and/or stale signatures.

## Implementation order
This is implementation part 1 of 2, which adds an optional property called `GeneratedFromTask` to the `SignRequest` model.

## Related Issue(s)
- #351
- Altinn/app-lib-dotnet#460

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Storage.Interface/Models/SignRequest.cs
+++ b/src/Storage.Interface/Models/SignRequest.cs
@@ -16,6 +16,12 @@ namespace Altinn.Platform.Storage.Interface.Models
         public string SignatureDocumentDataType { get; set; }
 
         /// <summary>
+        /// The task which should be linked to this signature
+        /// </summary>
+        [JsonProperty(PropertyName = "generatedFromTask")]
+        public string GeneratedFromTask { get; set; } = string.Empty;
+        
+        /// <summary>
         /// List of dataElementSignatures
         /// </summary>
         [JsonProperty(PropertyName = "dataElementSignatures")]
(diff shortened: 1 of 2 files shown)


src/Storage.Interface/Models/SignRequest.cs
```diff
@@ -15,6 +15,12 @@ public class SignRequest
         [JsonProperty(PropertyName = "signatureDocumentDataType")]
         public string SignatureDocumentDataType { get; set; }
 
+        /// <summary>
+        /// The task which should be linked to this signature
+        /// </summary>
+        [JsonProperty(PropertyName = "taskId")]
+        public string TaskId { get; set; } = string.Empty;
```
bjorntore, line 22: @danielskovli Probably a bit late to mention, but you can read the current taskId from this property: `instance.Process.CurrentTask?.ElementId`

danielskovli: Which begs the most important question: why is there no :facepalm: emoji available when reacting to comments?

danielskovli:
After discussing with the Core team, I've been informed that this PR needs to split in two:
- Alter the data model in `Storage.Interface`
- Make use of the new data model in `Storage`

This due to the way `Storage` brings in the interface models, via `PackageReference` (NuGet).

SandGrainOne (approved):

PR #397, opened 2024-05-06
#396 Fix for the issue #396

### LINKED ISSUE #396: Kall mot GET /storage/api/v1/instances?org=dmf&size=50  og header X-Ai-InstanceOwnerIdentifier filtrerer ikke resultatet
### Description of the bug

jeg forsøker å hente ut alle instanser for en part basert på org nr:
GET /storage/api/v1/instances?org=dmf&size=50

X-Ai-InstanceOwnerIdentifier: ***
Der jeg har forsøkt å erstatte *** med "<orgnr>" , og "Organisation:<orgnr>".

Jeg får alltid alle skjema fra _alle_ parter, men forventer å bare å få instanser fra _gitt org_. 

I dokumentasjonen står det beskrevet at med gitt header skal svaret være filtrert.
Screenshot 2024-05-03 at 10 14 25

Workaround er å laste ned alle instanser for alle apper, og så filtrere på klientsiden.

### Steps To Reproduce

Gjør et kall:
GET /storage/api/v1/instances?org=dmf

X-Ai-InstanceOwnerIdentifier: <orgnr>

### Additional Information

_No response_

codemascot:
Moved conditional later to fix #396

## Description
Moved conditional later to fix #396 

## Related Issue(s)
- #3 
- #396

## Verification
- [x] All tests run green

### DIFF
--- a/src/Storage/Controllers/InstancesController.cs
+++ b/src/Storage/Controllers/InstancesController.cs
@@ -182,28 +182,28 @@ namespace Altinn.Platform.Storage.Controllers
             }
             else if (userId != null)
             {
-                if (instanceOwnerPartyId == null)
+                if (instanceOwnerPartyId == null && string.IsNullOrEmpty(instanceOwnerIdentifier))
                 {
-                    if (string.IsNullOrEmpty(instanceOwnerIdentifier))
-                    {
-                        return BadRequest("Either InstanceOwnerPartyId or InstanceOwnerIdentifier need to be defined.");
-                    }
-
-                    (string instanceOwnerIdType, string instanceOwnerIdValue) = InstanceHelper.GetIdentifierFromInstanceOwnerIdentifier(instanceOwnerIdentifier);
-
-                    if (string.IsNullOrEmpty(instanceOwnerIdType) || string.IsNullOrEmpty(instanceOwnerIdValue))
-                    {
-                        return BadRequest("Invalid InstanceOwnerIdentifier.");
-                    }
-
-                    (string person, string orgNo) = InstanceHelper.SeparatePersonAndOrgNo(instanceOwnerIdType, instanceOwnerIdValue);
-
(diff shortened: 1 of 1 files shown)


src/Storage/Controllers/InstancesController.cs
```diff
@@ -141,51 +141,53 @@ public class InstancesController : ControllerBase
...
-                }
+                case (false, true):
+                    break;
+
+                default:
+                    return BadRequest();
             }
-            else
+
+            if (instanceOwnerPartyId == null)
```
SandGrainOne, line 174: Instance owner party id is required only for end users. If the caller is org, then instance owner is optional.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -141,51 +141,53 @@ public class InstancesController : ControllerBase
                 return BadRequest("Both InstanceOwner.PartyId and InstanceOwnerIdentifier cannot be present at the same time.");
             }
 
-            if (orgClaim != null)
+            switch ((orgClaim != null, userId != null))
```
acn-sbuad, line 144: I think the if statements are easier to make sense of compared to the switch. What motivated the change?

SandGrainOne, line 144: I'm not sure turning the if into a switch was necessary. I would revert this.

SandGrainOne: As we talked about. Instead of if-else do separate tests for different cases and exit early.

```
if (orgclaim is null and userid is null {
   return BadRequest()
}

If (orgClaim is not null) {
   validate and handle required parameters for org
}

Perform party "normalization"

If (user is not null) {
   validate and handle required parameters for user
}
```

acn-sbuad (approved):
🎈

Issue #399, opened 2024-05-07
Non existent party in header X-Ai-InstanceOwnerIdentifier results in internal server error


SandGrainOne:
### Description of the bug

The new input parameter (header) X-Ai-InstanceOwnerIdentifier doesn't handle cases where Register returns not found. It results in an internal server error.

If the register lookup returns 404, then the operation as a whole should return an empty list of instances. We do this to prevent Storage from being a place where an external party can clean national identity numbers.  The result should be the same for Not Found and No access.

### Steps To Reproduce

Tested in AT22:

1. Create a valid token for application owner ttd using token generator.
2. Craft a valid request where the value in X-Ai-InstanceOwnerIdentifier is a none existing organization number
    https://platform.at22.altinn.cloud:443/storage/api/v1/instances?org=ttd
    X-Ai-InstanceOwnerIdentifier=Organisation:312508728
3. Run the request and observe the error message.

SandGrainOne:
Tested ok.

PR #400, opened 2024-05-07
#399 Implementation of exception handling for exception thrown by register for wrong person/org number 

codemascot:
#399 Implementation of exception handling for exception thrown by register for wrong person/org number 

## Description
#399 Implementation of exception handling for exception thrown by register for wrong person/org number 

## Related Issue(s)
- #3 
- #399 
- #396 

## Verification
- [x] All tests run green

### DIFF
--- a/src/Storage/Controllers/InstancesController.cs
+++ b/src/Storage/Controllers/InstancesController.cs
@@ -185,7 +185,22 @@ namespace Altinn.Platform.Storage.Controllers
 
                 (string person, string orgNo) = InstanceHelper.SeparatePersonAndOrgNo(instanceOwnerIdType, instanceOwnerIdValue);
 
-                instanceOwnerPartyId = await _registerService.PartyLookup(person, orgNo);
+                try
+                {
+                    instanceOwnerPartyId = await _registerService.PartyLookup(person, orgNo);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "An error occurred during PartyLookup with either person number: {person} or organisation number: {orgNo}", person, orgNo);
+
+                    QueryResponse<Instance> response = new()
+                    {
+                        Instances = new List<Instance>(),
+                        Count = 0,
+                    };
+
+                    return Ok(response);
+                }
             }
 
             if (!string.IsNullOrEmpty(continuationToken))


src/Storage/Controllers/InstancesController.cs
```diff
@@ -185,7 +185,21 @@ public class InstancesController : ControllerBase
 
                 (string person, string orgNo) = InstanceHelper.SeparatePersonAndOrgNo(instanceOwnerIdType, instanceOwnerIdValue);
 
-                instanceOwnerPartyId = await _registerService.PartyLookup(person, orgNo);
+                try
+                {
```
acn-sbuad, line 189: Should we make a distinction between PlatformException because org / person was not found (404 response) and PlatformException because bridge is unavailable or timeout for instance?

acn-sbuad: With current state of code, what happens if it is not a PartyNotFoundException? What would be our preferred behavior here ?

src/Storage/Services/RegisterService.cs
```diff
@@ -103,6 +105,14 @@ public async Task<int> PartyLookup(string person, string orgNo)
                 Party party = await response.Content.ReadFromJsonAsync<Party>(_serializerOptions);
                 return party.PartyId;
             }
+            else if (response.StatusCode == HttpStatusCode.NotFound)
+            {
+                string reason = await response.Content.ReadAsStringAsync();
+                _logger.LogError("// RegisterService // PartyLookup // Party not found. Response status code is {StatusCode}. \n Reason {Reason}.", response.StatusCode, reason);
+
+                // Return a custom exception indicating that the party was not found
+                throw await PartyNotFoundException.CreateAsync(response);
```
SandGrainOne, line 114: It's tempting to simplify this down to `return null;`

This is not an error or "exceptional". It's going to happen somewhat regularly.

There are two possible reasons for logging this:
1. We need statistics. How often are we given a person or organization that doesn't exist.
2. We need to analyze a support case. Someone is trying to download instances on an organization that is unknown to Altinn.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -185,7 +186,24 @@ public class InstancesController : ControllerBase
...
+                    instanceOwnerPartyId = await _registerService.PartyLookup(person, orgNo);
+                }
+                catch (Exception ex)
```
SandGrainOne, line 193: If we decide to keep the exception I would like the catch to be specific.
```suggestion
                catch (PartyNotFoundException partyNotFoundException)
```

src/Storage/Services/RegisterService.cs
```diff
@@ -103,6 +105,13 @@ public async Task<int> PartyLookup(string person, string orgNo)
                 Party party = await response.Content.ReadFromJsonAsync<Party>(_serializerOptions);
                 return party.PartyId;
             }
+            else if (response.StatusCode == HttpStatusCode.NotFound)
+            {
+                string reason = await response.Content.ReadAsStringAsync();
+                _logger.LogWarning("// RegisterService // PartyLookup // Party not found. Response status code is {StatusCode}. \n Reason {Reason}.", response.StatusCode, reason);
```
SandGrainOne, line 111:
```suggestion
                _logger.LogWarning("// RegisterService // PartyLookup // Response status code is {StatusCode}.", response.StatusCode);
```

SandGrainOne (approved):

Issue #402, opened 2024-05-10
Providing invalid input in the X-Ai-InstanceOwnerIdentifier header results in an internal server error


SandGrainOne:
### Description of the bug

The new input parameter (header) X-Ai-InstanceOwnerIdentifier doesn't handle cases where Register returns bad request. It results in an internal server error.

If the register lookup returns 400, then the operation as a whole should return a Bad Request response. 

### Steps To Reproduce

Tested in AT22:
Endpoint: https://platform.at22.altinn.cloud/storage/api/v1/instances?org=ttd

1. Create a valid token for application owner ttd using token generator.
2. Craft a request where the party value in X-Ai-InstanceOwnerIdentifier is invalid . E.g
    1. X-Ai-InstanceOwnerIdentifier=organisation:test
    2. X-Ai-InstanceOwnerIdentifier=person:42
3. Run the request and observe the error message.

SandGrainOne:
Tested, approved

PR #403, opened 2024-05-13
#402 Fix for handeling 400 response from register

codemascot:
#402 Fix for handeling 400 response from register

## Description
#402 Fix for handeling 400 response from register

## Related Issue(s)
- #402

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Controllers/InstancesController.cs
+++ b/src/Storage/Controllers/InstancesController.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Web;
 
@@ -191,7 +192,7 @@ namespace Altinn.Platform.Storage.Controllers
                 {
                     if (partyType == PartyType.Person)
                     {
-                        if (instanceOwnerIdValue.Length != 11)
+                        if (!Regex.IsMatch(instanceOwnerIdValue, @"^\d{11}$"))
                         {
                             return BadRequest("Person number needs to be exactly 11 digits.");
                         }
@@ -200,7 +201,7 @@ namespace Altinn.Platform.Storage.Controllers
                     }
                     else if (partyType == PartyType.Organisation)
                     {
-                        if (instanceOwnerIdValue.Length != 9)
+                        if (!Regex.IsMatch(instanceOwnerIdValue, @"^\d{9}$"))
                         {
                             return BadRequest("Organisation number needs to be exactly 9 digits.");
                         }
--- a/src/Storage/Controllers/InstancesController.cs
+++ b/src/Storage/Controllers/InstancesController.cs
@@ -183,6 +183,11 @@ namespace Altinn.Platform.Storage.Controllers
                     return BadRequest("Invalid InstanceOwnerIdentifier.");
                 }
 
+                if (instanceOwnerIdValue.Length != 11)
+                {
+                    return BadRequest("InstanceOwnerIdentifier value needs to be exactly 11 digits.");
+                }
+
                 (string person, string orgNo) = InstanceHelper.SeparatePersonAndOrgNo(instanceOwnerIdType, instanceOwnerIdValue);
 
                 instanceOwnerPartyId = _registerService.PartyLookup(person, orgNo).GetAwaiter().GetResult();
--- a/src/Storage/Controllers/InstancesController.cs
+++ b/src/Storage/Controllers/InstancesController.cs
@@ -202,7 +202,7 @@ namespace Altinn.Platform.Storage.Controllers
                     {
                         if (instanceOwnerIdValue.Length != 9)
                         {
-                            return BadRequest("Organisation number needs to be exactly 8 or 9 digits.");
+                            return BadRequest("Organisation number needs to be exactly 9 digits.");
                         }
 
                         orgNo = instanceOwnerIdValue;
--- a/src/Storage/Controllers/InstancesController.cs
+++ b/src/Storage/Controllers/InstancesController.cs
@@ -192,7 +192,8 @@ namespace Altinn.Platform.Storage.Controllers
                 {
                     if (partyType == PartyType.Person)
                     {
-                        if (!Regex.IsMatch(instanceOwnerIdValue, @"^\d{11}$"))
+                        Regex regex = InstanceOwnerIdRegExHelper.ElevenDigitRegex();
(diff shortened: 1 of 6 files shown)


test/UnitTest/TestingControllers/InstancesControllerTests.cs
```diff
@@ -622,6 +623,87 @@ public async Task GetEmptyListOfInstances_UserRequestsInstancesNoPartyIdDefinedA
...
+            Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
+            Assert.Contains(expectedResponseMessage, responseMessage);
+        }
+
+        /// <summary>
+        /// Test case: Get Multiple instances with invalid organisation number and without specifying instance owner partyId.
+        /// Expected: Controller throws 400 bad request.
+        /// </summary>
+        [Fact]
+        public async Task GetMany_UserRequestsInstancesNoPartyIdDefinedAndWithInvalidPerson_ThrowsBadRequest()
```
acn-sbuad, line 686:
```suggestion
        public async Task GetMany_UserRequestsInstancesNoPartyIdDefinedAndWithInvalidPerson_ReturnsBadRequest()
```

No exception involved, use returns instead of throws

test/UnitTest/TestingControllers/InstancesControllerTests.cs
```diff
@@ -622,6 +623,87 @@ public async Task GetEmptyListOfInstances_UserRequestsInstancesNoPartyIdDefinedA
             registerService.VerifyAll();
         }
 
+        /// <summary>
+        /// Test case: Get Multiple instances with wrong organisation number and without specifying instance owner partyId.
+        /// Expected: Register service throws 400 bad request.
+        /// </summary>
+        [Fact]
+        public async Task GetMany_UserRequestsInstancesNoPartyIdDefinedAndWithWrongOrganisation_ThrowsBadRequest()
+        {
```
acn-sbuad, line 632: A bit confused by this test. orgno is 9 digits and should be valid based on the error message. So not sure what "wrong" org number means  here?

codemascot: Spot on! I agree this test is misleading because of incorrect message and uses of "wrong". 👍

src/Storage/Controllers/InstancesController.cs
```diff
@@ -183,7 +184,30 @@ public class InstancesController : ControllerBase
...
+                            return BadRequest("Person number needs to be exactly 11 digits.");
+                        }
+
+                        person = instanceOwnerIdValue;
+                    }
+                    else if (partyType == PartyType.Organisation)
+                    {
+                        if (instanceOwnerIdValue.Length != 9)
+                        {
+                            return BadRequest("Organisation number needs to be exactly 8 or 9 digits.");
```
SandGrainOne, line 205:
```suggestion
                            return BadRequest("Organisation number needs to be exactly 9 digits.");
```

test/UnitTest/TestingControllers/InstancesControllerTests.cs
```diff
@@ -622,6 +623,88 @@ public async Task GetEmptyListOfInstances_UserRequestsInstancesNoPartyIdDefinedA
             registerService.VerifyAll();
         }
 
+        /// <summary>
+        /// Test case: Get Multiple instances with an organisation number(which for some reason ended up in bad request in the register) and without specifying instance owner partyId.
+        /// Expected: Register service throws 400 bad request.
+        /// </summary>
+        [Fact]
+        public async Task GetMany_UserRequestsInstancesNoPartyIdDefinedAndWithOrganisation_ThrowsBadRequest()
```
acn-sbuad, line 631: Returns, not throws.
I would probably rename to GetMany_RegisterServiceThrowsBadRequestException_ReturnsBadRequest.

codemascot: As the controller is returning bad request and this is a test for controller then it should be returns.

src/Storage/Services/RegisterService.cs
```diff
@@ -111,11 +111,15 @@ public async Task<int> PartyLookup(string person, string orgNo)
 
                 return -1;
             }
-            else
+            else if (response.StatusCode == HttpStatusCode.BadRequest)
```
acn-sbuad, line 114: not a buge fan of this exception throwing. Now that validation of input has been added to the controller we should not be getting bad requests. And if we do, it can be handled like any platform exception.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -183,7 +184,30 @@ public class InstancesController : ControllerBase
...
 
-                (string person, string orgNo) = InstanceHelper.SeparatePersonAndOrgNo(instanceOwnerIdType, instanceOwnerIdValue);
+                string person = null;
+                string orgNo = null;
+
+                if (Enum.TryParse<PartyType>(instanceOwnerIdType, true, out PartyType partyType))
+                {
+                    if (partyType == PartyType.Person)
+                    {
+                        if (instanceOwnerIdValue.Length != 11)
```
SandGrainOne, line 194: We might want to add a regex here. To ensure that it's 11 digits and not a mix off digits and letters. The current check doesn't catch `hhhhhhhhhhh` for example. It would fall through and we would have an unnecessary call to Register.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -183,7 +184,30 @@ public class InstancesController : ControllerBase
...
+                        {
+                            return BadRequest("Person number needs to be exactly 11 digits.");
+                        }
+
+                        person = instanceOwnerIdValue;
+                    }
+                    else if (partyType == PartyType.Organisation)
+                    {
+                        if (instanceOwnerIdValue.Length != 9)
```
SandGrainOne, line 203: Same as for Person it's now possible to get past the check with an invalid organization number by including letters.

src/Storage/Services/RegisterService.cs
SandGrainOne: My initial thought were that any BadRequest response we get from Register should result in a BadRequest response from Storage, but that might not be the case. I'm now inclined to agree with @acn-sbuad

codemascot: Yep! I was thinking about the same as well, which actually makes this test obsolete! 👍 

https://github.com/Altinn/altinn-storage/pull/403#discussion_r1601382755

acn-sbuad (approved):
🎈

PR #413, opened 2024-05-29
Add Metadata to DataElement [part 1/2].

bjorntore:
DIBK has requested support for adding custom metadata to data elements.
https://github.com/Altinn/app-lib-dotnet/issues/392

## Description
They wanted something similar to "ekstraMetadata" here: https://developers.fiks.ks.no/tjenester/svarut/integrasjon/restv1/#dokumentmetadata

Considered using object for value, but they though string was good.
Should KeyValueEntry be in it's own file, or fine in the same file?

**There will be another PR with the required changes to set this property.**

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/392

## Verification
- [] **Your** code builds clean without any errors or warnings
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Storage.Interface/Models/DataElement.cs
+++ b/src/Storage.Interface/Models/DataElement.cs
@@ -96,7 +96,7 @@ namespace Altinn.Platform.Storage.Interface.Models
         public List<string> Tags { get; set; } = new List<string>();
 
         /// <summary>
-        /// Gets or sets a list of key-value pairs that can be used to store metadata.
+        /// Gets or sets a list of key-value pairs that can be used to store metadata. Can be used in custom app code to store custom metadata.
         /// </summary>
         [JsonProperty(PropertyName = "metadata")]
         public List<KeyValueEntry> Metadata { get; set; }
--- a/src/Storage.Interface/Models/DataElement.cs
+++ b/src/Storage.Interface/Models/DataElement.cs
@@ -96,6 +96,12 @@ namespace Altinn.Platform.Storage.Interface.Models
         public List<string> Tags { get; set; } = new List<string>();
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Storage.Interface/Models/DataElement.cs
```diff
@@ -95,6 +95,12 @@ public class DataElement : ChangableElement
         [JsonProperty(PropertyName = "tags")]
         public List<string> Tags { get; set; } = new List<string>();
 
+        /// <summary>
+        /// Gets or sets a list of key-value pairs that can be used to store custom metadata.
+        /// </summary>
+        [JsonProperty(PropertyName = "customMetadata")]
+        public List<KeyValueEntry> CustomMetadata { get; set; } = new();
```
acn-sbuad, line 102: Is it relevant for all data elements or could it be nullable?

bjorntore: I think it will be quite niche, and could probably be nullable. Took inspiration from the Tags prop just above it when I decided to set it to new();

bjorntore: Do you think I should change it, or is it OK that it's auto initialized as an empty list?

ivarne: Storage.Interfaces is netstandard2.0 and does not support nullable reference types.

I'm still in favor of not initializing but adding a comment explaining the decision.

bjorntore: Will do that.

acn-sbuad (approved):
🎈

HauklandJ (approved):

PR #414, opened 2024-05-29
Fix `AbandonCurrentMoveToNext` authentication error for signature rejection

danielskovli:
## Description
Starting with [Altinn.App.Core v8.2.0](https://github.com/Altinn/app-lib-dotnet/releases/tag/v8.2.0), the reject action in a signature step produces a `FlowType=AbandonCurrentMoveToNext` instead of `FlowType=CompleteCurrentMoveToNext`.

This in turn has led to a 403 Forbidden error from the storage service due to incorrect permissions ([see issue](https://github.com/Altinn/app-lib-dotnet/issues/660)).

The code change in this PR fixes the issue _in this very specific case_, but I am uncertain of other potential side-effects of this change. The [ProcessControllerTest](https://github.com/Altinn/altinn-storage/blob/29d3b42586160dbb7e7da2ff817155205de5272f/test/UnitTest/TestingControllers/ProcessControllerTest.cs) runs green, but it's unclear to me if this covers all necessary scenarios. 

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/660

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Controllers/ProcessController.cs
+++ b/src/Storage/Controllers/ProcessController.cs
@@ -84,7 +84,7 @@ namespace Altinn.Platform.Storage.Controllers
             string taskId = null;
 
             var moveNextFlows = new[] { "CompleteCurrentMoveToNext", "AbandonCurrentMoveToNext" };
-            if (processState?.CurrentTask?.FlowType != null && !moveNextFlows.Contains(processState.CurrentTask.FlowType))
+            if (processState?.CurrentTask?.FlowType is not null && !moveNextFlows.Contains(processState.CurrentTask.FlowType))
             {
                 altinnTaskType = processState.CurrentTask.AltinnTaskType;
                 taskId = processState.CurrentTask.ElementId;
--- a/src/Storage/Controllers/ProcessController.cs
+++ b/src/Storage/Controllers/ProcessController.cs
@@ -83,7 +83,8 @@ namespace Altinn.Platform.Storage.Controllers
             string altinnTaskType = existingInstance.Process?.CurrentTask?.AltinnTaskType;
             string taskId = null;
 
-            if (processState?.CurrentTask?.FlowType != null && !processState.CurrentTask.FlowType.Equals("CompleteCurrentMoveToNext"))
+            var moveNextFlows = new[] { "CompleteCurrentMoveToNext", "AbandonCurrentMoveToNext" };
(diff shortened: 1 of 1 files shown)


src/Storage/Controllers/ProcessController.cs
```diff
@@ -83,7 +83,8 @@ public class ProcessController : ControllerBase
             string altinnTaskType = existingInstance.Process?.CurrentTask?.AltinnTaskType;
             string taskId = null;
 
-            if (processState?.CurrentTask?.FlowType != null && !processState.CurrentTask.FlowType.Equals("CompleteCurrentMoveToNext"))
+            var moveNextFlows = new[] { "CompleteCurrentMoveToNext", "AbandonCurrentMoveToNext" };
+            if (processState?.CurrentTask?.FlowType != null && !moveNextFlows.Contains(processState.CurrentTask.FlowType))
```
HauklandJ, line 87: Consider using `is not null` instead of `!= null`. The only difference (besides the syntax) is, that the compiler guarantees that no user-overloaded operator is called when using `is not null` instead of `!= nul`l (or `is null `instead of `== null`).

danielskovli: Completely agree. I guess I took the "minimal PR changes" concept a step too far 🤷. I'll make the change now.

danielskovli:
@SandGrainOne / @acn-sbuad : If you agree with these changes, can you please merge, publish and deploy a new version whenever your next deployment window is?

The timeline is not critical, but if you have an ETA at some point I will report back to the person who reported the bug.

SandGrainOne (approved):

SandGrainOne:
I honestly have no idea either. I'm going to trust you to do some serious regression testing. Next ordinary TT02 deploy is on Wednesday.

danielskovli:
> I'm going to trust you to do some serious regression testing

I'll do what I can. The other [commenters on the issue](https://github.com/Altinn/app-lib-dotnet/issues/660#issuecomment-2137427227) are also in favour of the fix. Not much more to be done before deploy at this stage, I think.

SandGrainOne:
The change can be tested in AT. No need to wait for Wednesday deploy.
You should be allowed to merge yourself now that the PR has been approved.

danielskovli:
I'm not sure how and what testing in AT entails, but challenge accepted. Leave it with me.

PR #426, opened 2024-06-03
`DisallowUserInstantiation` property on `Application` to prevent manual instantiation of instances

martinothamar:
## Description
We need configuration to block "manual instantiation" by users, for applications where service owner
create and prefill instances and then let users fill in the rest.

Changes
* `InstantiationConfig.ManualInstantiationDisabled` to disable manual instantiation (not by service owner)
* Docs improvements on `PartyTypesAllowed`
* EditorConfig fixes as it seems all files have LF line endings and not CRLF (my editor was trying to correct line endings all the time)

Lib PR: https://github.com/Altinn/app-lib-dotnet/pull/671

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/361

## Verification
- [ ] All tests run green

## Documentation


src/Storage.Interface/Models/PartyTypesAllowed.cs
```diff
@@ -1,36 +1,76 @@
...
+        /// Gets or sets a value indicating whether a sub unit is allowed to be the owner of an instance.
         /// </summary>
         [JsonProperty(PropertyName = "subUnit")]
         public bool SubUnit { get; set; }
+
+        /// <summary>
+        /// Gets or sets the list of allowed instantiator types for an application.
+        /// </summary>
+        [JsonProperty(PropertyName = "allowedInstantiatorPartyTypes")]
+        public List<InstantiatorPartyType> AllowedInstantiatorPartyTypes { get; set; }
```
HauklandJ, line 42: Consider renaming
PartyTypesAllowed.AllowedInstantiatorPartyTypes -> PartyTypesAllowed.Instantiators,
PartyTypesAllowed.AllowedCopyInstantiatorPartyTypes  -> PartyTypesAllowed.CopyInstantiators?

src/Storage.Interface/Models/PartyTypesAllowed.cs
```diff
@@ -1,36 +1,76 @@
...
+        /// Gets or sets the list of allowed instantiator types for an application when instantiation is done by copy.
+        /// </summary>
+        [JsonProperty(PropertyName = "allowedCopyInstantiatorPartyTypes")]
+        public List<InstantiatorPartyType> AllowedCopyInstantiatorPartyTypes { get; set; }
+    }
+
+    /// <summary>
+    /// Enum representing the different types of instantiators.
+    /// </summary>
+    public enum InstantiatorPartyType
```
ivarne, line 54: enums are written as numbers with System.Text.Json, so we need some annotations (or just not use enums)

martinothamar:
This PR (currently) is just for exploring options, so no need to review specifics of the code just yet. Location, semantics, overlap with existing config etc is the stuff where I don't have enough context to know what's a good idea or not, updating the description

HauklandJ:
> This PR (currently) is just for exploring options, so no need to review specifics of the code just yet. Location, semantics, overlap with existing config etc is the stuff where I don't have enough context to know what's a good idea or not, updating the description

As `instantiation` is not a party type, but rather the context in which a party type exists, it may be confusing to expand PartyTypesAllowed. Maybe it could be renamed to something that makes sense, but I fail to find an umbrella term thats easy to interpret as encompassing parties, instantiation and copy instantiation (the latter of which I do not fully understand what is represented with). 

With this in mind, I vote to keep it as seperate config.

martinothamar:
Updated the PR commit and description with what we agreed yesterday

martinothamar:
@altinnadmin it would be nice if you could review this as well from a domain/naming POV 😄

HauklandJ (approved):
lgtm

altinnadmin:
- "Manual instantiation" is not a concept, and I don´t think its wise to introduce it
- Instantiation is an action in XACML and API in the app, and the app owner currently configures who is allowed to trigger an instantiation. If only the app owner should be allowed to instantiate, it should be configured as an authorization rule.
- If a user should be allowed to instantiate only when creating a copy, I guess the "clean" way to do that is to introduce a new copy-action to be used in XACML, since these are different concepts that needs different authorization rules?

martinothamar:
We discussesd this again - XACML based config for this is decided against for practical reasons - changes needed to accomodate this would affect different services such as - authorization, storage, backend/lib. There may also be more dimensions/config to this in future versions

Updated proposal for applicationmetadata:

```csharp
        /// <summary>
        /// Gets or sets the list of allowed instantiators for this app, and limits it to orgs/service owners.
        /// The "urn:altinn:org" claim of the current identity must match one of the values in this list.
        /// If the list is null, any org/service owner or user can instantiate.
        /// </summary>
        [JsonProperty(PropertyName = "instantiationAllowedBy")]
        public List<string> InstantiationAllowedBy { get; set; }
```

martinothamar:
Since XACML already does auth on which orgs can instantiate, I've made the configuration parameter a simple bool again

PR #432, opened 2024-06-13
Add CustomMetadata to DataElement.

bjorntore:
After discussing with internally and with DIBK we have decided that it would be useful to have two different metadata properties on data element.

1. One that is controlled by the end user, for custom metadata sent in through an endpoint in the altinn app.
2. One that is controlled by the application developer, and that can not be changed by the end user. Set through custom backend code.

I could do fine with just the first one for now, but the naming of the "Metadata" property that is currently merged into main in Storage.Interface matches best with the second variant, and I probably can't rename that property to match the first variant better now that it's out in main? @acn-sbuad 

**Very open to better name suggestions.**

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/392


acn-sbuad (approved):
🎈

Issue #439, opened 2024-06-19
Changes needed for migration


HenningNormann:
# Changes needed for migration
## App
- Introduce external model vs internal model
- Expand internal (and external?) model
    - Version number (e.g. 2 and 3 for a2 and a3)
    - _Blob storage root path (done)_
- _Application need info about xsls and their internal order (done)_
- _Code lists referenced from the xsls (done)_

## Instance/instance event/data element
- Introduce external model vs internal model
- Expand model
     - Version number (e.g. 2 and 3 for a2 and a3)
     - _More instanceevent types? (Done)_
     - _PresentationText also in DataElement (Done with the metadataproperty)_
- _Support for generating a data element's content on the fly and connect a module to do the generating (Done)_
- Hide migrated a3 instances from sbl search
- _Support for multiple blobs (and/or accounts) for each service owner (done)_

## _Temporary (Done)_
- _New controller to upload instances, ievents, data elements and applications (Done)_
- _Table to store migration state (one row for each a3 service: a2 archive reference and a3 instance guid) (Done)_

## _Misc (Done)_
- _How to support presentationtexts on the data element level. Current thoughts:_
    - _Expand internal model for DataElement with property for PresentationTexts_
    - _Migration program_
        - _Generate main form values into instance (used for archive title)_
        - _Generate values for each form into each data element (not used initially, but to avoid to lose significant data)_
        - _Generate filenames that includes the values for each form_

PR #489, opened 2024-09-10
A2 migration changes 6b - support for pdf landscape/portrait

HenningNormann:
## Description
- Pdf generation changes
    - Property IsPortrait for each style sheet - including database and api
    - Set config on api call to browserless/puppeteer
    - Generate pdf for each page at a time and merge (with PDFsharp) result to single pdf when form has a mix of landscape and portrait
    - Avoid page break after last page in form
    - Remove setting of page format (landscape vs portrait) in html
- Added method in MigrationController to test pdf generation

## Related Issue(s)
- #439 

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Controllers/ContentOnDemandController.cs
SandGrainOne, line 1: Could you clean up in warnings related to async await?

src/Storage/Migration/v0.12/01-setup-tables.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE storage.a2xsls ADD COLUMN IF NOT EXISTS isportrait BOOL NOT NULL DEFAULT true;
+DROP FUNCTION IF EXISTS storage.reada2xsls(_org TEXT, _app TEXT, _lformid INT, _language TEXT, _xsltype INT);
```
SandGrainOne, line 2: This is not backwards compatible. Wondering if `CREATE OR REPLACE` is good enough or if we need a new version.

src/Storage/Controllers/MigrationController.cs
```diff
@@ -415,14 +426,14 @@ public async Task<ActionResult> CreateImage()
         /// <param name="xsltype">Xsl type</param>
         /// <returns>Ok</returns>
         [AllowAnonymous]
-        [HttpPost("xsl/{org}/{app}/{lformid}/{pagenumber}/{language}/{xsltype}")]
+        [HttpPost("xsl/{org}/{app}/{lformid}/{pagenumber}/{language}/{xsltype}/{isportrait}")]
         [DisableFormValueModelBinding]
         [ProducesResponseType(StatusCodes.Status201Created)]
         [Produces("application/json")]
-        public async Task<ActionResult> CreateXsl([FromRoute] string org, [FromRoute] string app, [FromRoute] int lformid, [FromRoute] int pagenumber, [FromRoute] string language, [FromRoute] int xsltype)
+        public async Task<ActionResult> CreateXsl([FromRoute] string org, [FromRoute] string app, [FromRoute] int lformid, [FromRoute] int pagenumber, [FromRoute] string language, [FromRoute] int xsltype, [FromRoute(Name = "isportrait")] bool isPortrait)
```
SandGrainOne, line 433: Update documentation to clear warnings.

src/Storage/Migration/v0.12/01-setup-tables.sql
HenningNormann: Backwards compability is not an issue because this PR should be released long before it affects other than team dataflytt members.

src/Storage/Controllers/ContentOnDemandController.cs
```diff
@@ -122,7 +124,61 @@ public async Task<Stream> GetPaymentAsHtml([FromRoute] string org, [FromRoute] s
...
+                    MemoryStream mergedPdf = new MemoryStream();
+                    mergedDoc.Save(mergedPdf);
+                    return mergedPdf;
+                }
+            }
+            else
+            {
+                // Generate all pages in a single operation
+                return await _pdfGeneratorClient.GeneratePdf(
+                    $"{Request.Scheme}://{Request.Host}{Request.PathBase}{Request.Path}{Request.QueryString}".Replace("formdatapdf", "formdatahtml"),
```
SandGrainOne, line 179: Could it be possible for a crafty hacker to make a specially made request that we now relatively blindly forward to puppeteer? Any dangers with this and should we build the URL based on our knowledge of the path and configuration instead?

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: Done

src/Storage/Controllers/MigrationController.cs
HenningNormann: Done

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: The ContentOnDemandController is only accessible from inside the cluster. If a hacker gets access, the main problem is the data the hacker is able to get in return - not the url that is sent to puppeteer.

SandGrainOne: Is it unnecessary to inldude QueryString?

src/Storage/Migration/v0.12/01-setup-tables.sql
SandGrainOne: Still not sure why we have DROP FUNCTION here when we have CREATE OR REPLACE later.

SandGrainOne (approved):

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: No, you're right. It's removed now.

src/Storage/Migration/v0.12/01-setup-tables.sql
HenningNormann: It's just to remove the old version with fewer parameters - REPLACE doesn't touch the old one. Normally (when it affects real users) we must postpone the delete until the next release, and it's normally forgotten.

PR #539, opened 2024-10-31
Changed  a2 pdf generation to push html to the generator

HenningNormann:
## Description
- Changed pdf generation to push html to the generator
- Avoid multiple convertions between string and stream when generating pdf
- Bumped storage.interface to 4.0.3

## Related Issue(s)
- #439

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Controllers/ContentOnDemandController.cs
+++ b/src/Storage/Controllers/ContentOnDemandController.cs
@@ -2,8 +2,10 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 using Altinn.Platform.Storage.Clients;
 using Altinn.Platform.Storage.Configuration;
 using Altinn.Platform.Storage.Interface.Models;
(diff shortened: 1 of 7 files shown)


src/Storage/Controllers/ContentOnDemandController.cs
```diff
@@ -204,10 +203,15 @@ public async Task<Stream> GetFormdataAsHtml([FromRoute]string org, [FromRoute] s
...
         public async Task<Stream> GetFormSummaryAsHtml([FromRoute] string org, [FromRoute] string app, [FromRoute] Guid instanceGuid, [FromRoute] Guid dataGuid, [FromRoute] string language)
         {
-            return await GetFormdataAsHtmlInternal(org, app, instanceGuid, dataGuid, language, 2);
+            return await GetFormdataAsHtmlStream(app, instanceGuid, dataGuid, language, 2);
         }
 
-        private async Task<Stream> GetFormdataAsHtmlInternal(string org, string app, Guid instanceGuid, Guid dataGuid, string language, int viewType, int singlePageNr = -1)
+        private async Task<Stream> GetFormdataAsHtmlStream(string app, Guid instanceGuid, Guid dataGuid, string language, int viewType, int singlePageNr = -1)
+        {
+            return new MemoryStream(Encoding.UTF8.GetBytes(await GetFormdataAsHtmlString(app, instanceGuid, dataGuid, language, viewType, singlePageNr)));
```
SandGrainOne, line 211: I'm surprised it's not reversed. That the base method produce a stream and not a string, and then to get a string you read the stream.

SandGrainOne (approved):

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: The base method is cut and paste from a2. All the logic is based on a string. If this method should produce a stream, we would introduce a back and forth conversion for the most used use case.

PR #551, opened 2024-11-14
Endpoint for process state and events batch update

martinothamar:
## Description

Endpoint for upating process state and instsanceevents in a batch

## Related Issue(s)
- #544 

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

Issue #544, opened 2024-11-08
Update process state and history (`InstanceEvents`) within a single transaction


martinothamar:
### Description

During `process/next` in an app, when transitional work has been completed and we are ready to commit the new state and events to Storage, we send

* 1 request for the process state update
* 1 request for each `InstanceEvent` generated

This leads to a race condition for external systems that are polling the Storage API for new instances to pull down. In particular it was noted that when there network issues related to a specific app cluster, the window for the race condition widened to the point that alot of the instances that were polled had updated process state (they were completed) yet not all `InstanceEvent`s were stored. We should have an API for updating both the process state and the `InstanceEvent`s in a single transaction to avoid this issue.

### Additional Information

A workaround for polling systems is to only consume instances that have been completed for some time (importantt to note that this only reduces the probability of errors states)

### Tasks

- [x] Add API to altinn-storage
- [x] Implement API in app-lib, merge and release when new Storage version is deployed 
- [x] Update localtest to match Storage

### Acceptance Criterias

* Apps on the newest version of the backend libriaries use the new endpoint which posts process state update and instance events in a single request, and get back the updated instance
* Storage runs this request in a single transaction
* A new repository is implemented in Storage

martinothamar:
Just discused with @bjorntore, some uncertainty on what we should do about the API contract itself. Ideally we don't want to have Storage authorizing this API request as we have already done so in the app. It is tempting to create a v2 of the API where we do authorization based on Platform Access Token for the APIs that are supposed to always come from the app. If we do Platform Access Token on an adjacent API in v1 things are starting to feel a little inconsistent and messy. What do you think @SandGrainOne ?

SandGrainOne:
We have as one of our tenants to authorize access in every layer. We might discuss exceptions when we implement "system activities", but in this case it's essencially a user action that should involve an authorization check of the user.

SimenRokaas:
Hei, er det noen oppdatering på dette? Dette er fortsatt et ganske kritisk problem for oss (DSB). Vi har vært nødt til å implementere en daglig sjekk på ukvitterte instanser, og må håndtere disse manuelt siden det vanlige mottaket vårt baserer seg på at det finnes et event hos dere.

25/11 meldte SSB og Patentstyret at de også hadde problemer med dette, ref https://altinn.slack.com/archives/C068VDDDR6W/p1732521571002409

martinothamar:
[...]

PR #562, opened 2024-11-25
Endpoint for process state and events in one transaction

martinothamar:
## Description
Replaces #551

## Related Issue(s)
- https://github.com/Altinn/altinn-storage/issues/544

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Repository/PgInstanceAndEventsRepository.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            instance = await reader.GetFieldValueAsync<Instance>("updatedInstance");
+        }
+
+        instance.Data = dataElements; // TODO: requery instead?
+
+        return PgInstanceRepository.ToExternal(instance);
+    }
+
+    private sealed record InstanceEventRecord(Guid Instance, Guid? AlternateId, InstanceEvent Event);
```
HenningNormann, line 82: Why this new record? Why not sending List<InstanceEvent> to the procedure?

martinothamar: I just made it as similar to `storage.insertInstanceevent`/`InsertInstanceEvent` as I could. I'm fine with just using `InstanceEvent`, but then I would have to do the InstanceId string splitting for the first column in psql instead right?

martinothamar: Sounds good to me, done that now

src/Storage/Migration/FunctionsAndProcedures/insertinstanceevents.sql
```diff
@@ -0,0 +1,11 @@
+
+CREATE OR REPLACE PROCEDURE storage.insertinstanceevents(_instance UUID, _events JSONB)
+    LANGUAGE 'plpgsql'	
+AS $BODY$
+BEGIN
+	-- Dummy comment to verify that new migration solution works end to end
```
HenningNormann, line 6: Remove dummy content

HenningNormann (approved):

src/Storage/Controllers/ProcessController.cs
```diff
@@ -80,28 +135,57 @@ public async Task<ActionResult<Instance>> PutProcess(
...
+        
+        /// <summary>
+        /// Updates the process state of an instance.
+        /// </summary>
+        /// <param name="instanceOwnerPartyId">The party id of the instance owner.</param>
+        /// <param name="instanceGuid">The id of the instance that should have its process updated.</param>
+        /// <param name="processStateUpdate">The new process state of the instance (including instance events).</param>
+        /// <returns></returns>
+        [Authorize]
+        [HttpPut("batch")]
```
martinothamar, line 168: Are we happy with the API name? Struggled to come up with alternatives here. It is a batch in the sense that there are multiple things (and multiple events). If the principle of modelling was behavior, and not resource/entity, it could be something like `CommitProcessTransition` or something like that, but that's not the case here..

Repo name is also less then ideal, but internal and easily changed.

cc @bjorntore @SandGrainOne

bjorntore: I would prefer something other than "batch" if it was up to me.

When I see batch I'm thinking I can supply a list of multiple things that should be processed in bulk. This endpoint seems to me as a pretty normal PUT operation that changes some data related to an instance. My idea of a batch might not be very nuanced though.

Hard to come up with good suggestions. 🤔 

I think I'm leaning towards keeping same URL as old endpoint, just with v2: 
`storage/api/v2/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/process`

Other suggestions? @SandGrainOne

martinothamar: So, the options are:
* We include all current APIs in a v2, where this one endpoint changes. We risk making a lot of different API versions over time if we set this precedent
* We include only this changed API in a v2. What will become the new precedent? New APIs - will these be added to v1? v2? Both? Changed APIs - add only to  latest (v2)? 
* Find a decent name for the new endpoint and sticking with v1

I'm a fan of sticking to v1 here becuase building a strategy for API versioning is more work then I'd like to take on atm. It's probably something that would have to be discussed at the architecture forums as ideally all Altinn 3 teams have a cohesive strategy for public APIs (which includes how to do versioning)

SandGrainOne: We can use a name similar to the naming of the new repository class: `instanceandevents`
```suggestion
        [HttpPut("instanceandevents")]
```

SandGrainOne (changes requested):
Rename endpoint

SandGrainOne (approved):

PR #564, opened 2024-11-25
fix: remove explicit org check in instancesController

mikaelrss:
This PR removes an explicit check that the organisation used in authentication is the same organisation that owns the app when requesting active instances in `InstancesController`. 

## Description
The reason we want to implement this change, is because in our procjet, Digital gravferd, we need to manually handle instance creation and deletion. To do this we need to be able to get active instances. Currently, we're not able to do this because we are using `digdir` as maskinporten tenant, where as the apps we want to manage instances for are hosted by the service owner `sfvt`. 

There are financial and organizational reasons for why we're using the `digdir` tenant, and we're investigating whether we can change this to use `sfvt` as maskinporten tenant. In the meantime, we still see this as a reasonable change, as who should have access to list active instances should be able to configure via the authorization policy.

This issue has been discussed with @SandGrainOne and the Apps team in a meeting previous to creating this PR.

## Verification
- [x] All tests run green


src/Storage/Controllers/InstancesController.cs
```diff
@@ -118,13 +118,59 @@ public async Task<ActionResult<QueryResponse<Instance>>> GetInstances(InstanceQu
...
+                
                 if (!orgClaim.Equals(queryParameters.Org, StringComparison.InvariantCultureIgnoreCase))
                 {
-                    return Forbid();
-                }
+                    var appId = ValidateAppId(queryParameters.AppId).Split("/")[1];
+                    XacmlJsonRequestRoot request;
+                    try
+                    {
+                        var instanceOwnerPartyId = (int)queryParameters.InstanceOwnerPartyId;
```
SandGrainOne, line 128: Instance owner will usually be undefined when the user is an application owner (org).

test/UnitTest/data/apps/sfvt/test-read-app-no-permission/config/authorization/policy.xml
```diff
@@ -0,0 +1,38 @@
...
+<xacml:Policy xmlns:xsl="http://www.w3.org/2001/XMLSchema-instance" xmlns:xacml="urn:oasis:names:tc:xacml:3.0:core:schema:wd-17" PolicyId="urn:altinn:example:policyid:1" Version="1.0" RuleCombiningAlgId="urn:oasis:names:tc:xacml:3.0:rule-combining-algorithm:deny-overrides">
+    <xacml:Target/>
+    <xacml:Rule RuleId="urn:altinn:example:ruleid:1" Effect="Permit">
+        <xacml:Description>A rule giving sfvt the right to read, write, complete and instantiate a given app of sfvt/dgm-signering</xacml:Description>
+        <xacml:Target>
+            <xacml:AnyOf>
+                <xacml:AllOf>
+                    <xacml:Match MatchId="urn:oasis:names:tc:xacml:1.0:function:string-equal">
+                        <xacml:AttributeValue DataType="http://www.w3.org/2001/XMLSchema#string">sfvt</xacml:AttributeValue>
+                        <xacml:AttributeDesignator AttributeId="urn:altinn:org" Category="urn:oasis:names:tc:xacml:3.0:attribute-category:resource" DataType="http://www.w3.org/2001/XMLSchema#string" MustBePresent="false" />
```
SandGrainOne, line 11: Category needs to be `urn:oasis:names:tc:xacml:1.0:subject-category:access-subject`, please see one of the existing policy files.

test/UnitTest/data/apps/sfvt/test-read-app-no-permission/config/authorization/policy.xml
```diff
@@ -0,0 +1,38 @@
...
+                    </xacml:Match>
+                    <xacml:Match MatchId="urn:oasis:names:tc:xacml:1.0:function:string-equal">
+                        <xacml:AttributeValue DataType="http://www.w3.org/2001/XMLSchema#string">test-read-app</xacml:AttributeValue>
+                        <xacml:AttributeDesignator AttributeId="urn:altinn:app" Category="urn:oasis:names:tc:xacml:3.0:attribute-category:resource" DataType="http://www.w3.org/2001/XMLSchema#string" MustBePresent="false" />
+                    </xacml:Match>
+                </xacml:AllOf>
+            </xacml:AnyOf>
```
SandGrainOne, line 18: This is the resource category. We also need a section for access-subject. Look at one of the existing policy files.

Missing section for Category `urn:oasis:names:tc:xacml:1.0:subject-category:access-subject`

SandGrainOne: The file is in a folder with name `test-read-app-no-permission`. Resource inside the file is `test-read-app`.

The tests are using a simple mocked authorization component and it's using naming conventions to find correct policy file (I think). Make sure the resource inside the file match the rest.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -118,13 +118,59 @@ public async Task<ActionResult<QueryResponse<Instance>>> GetInstances(InstanceQu
...
+                        request = DecisionHelper.CreateDecisionRequest(queryParameters.Org, appId, HttpContext.User, "read", instanceOwnerPartyId, null);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogError(ex, $"Something went wrong during CreateDecisionRequest for application id: {appId}");
```
SandGrainOne, line 133: This type of logging followed by a throw should be unnecessary.

If we need to add information to an exception, like a better message, then create a new exception with situational specific message and properties, the original exception can be provided as an inner exception.

Logging would be required if we didn't have the throw.

test/UnitTest/data/apps/sfvt/test-read-app-no-permission/config/authorization/policy.xml
```diff
@@ -0,0 +1,46 @@
...
+<xacml:Policy xmlns:xsl="http://www.w3.org/2001/XMLSchema-instance" xmlns:xacml="urn:oasis:names:tc:xacml:3.0:core:schema:wd-17" PolicyId="urn:altinn:example:policyid:1" Version="1.0" RuleCombiningAlgId="urn:oasis:names:tc:xacml:3.0:rule-combining-algorithm:deny-overrides">
+    <xacml:Target/>
+    <xacml:Rule RuleId="urn:altinn:example:ruleid:1" Effect="Permit">
+        <xacml:Description>A rule giving sfvt the right to read, write, complete and instantiate a given app of sfvt/dgm-signering</xacml:Description>
+        <xacml:Target>
+            <xacml:AnyOf>
+                <xacml:AllOf>
+                    <xacml:Match MatchId="urn:oasis:names:tc:xacml:3.0:function:string-equal-ignore-case">
+                        <xacml:AttributeValue DataType="http://www.w3.org/2001/XMLSchema#string">sfvt</xacml:AttributeValue>
+                        <xacml:AttributeDesignator AttributeId="urn:altinn:rolecode" Category="urn:oasis:names:tc:xacml:1.0:subject-category:access-subject" DataType="http://www.w3.org/2001/XMLSchema#string" MustBePresent="false" />
```
SandGrainOne, line 11: The value is not a role code, but a service owner code. `urn:altinn:org`

src/Storage/Controllers/InstancesController.cs
mikaelrss: Alright. So it should be fine to just remove these log statements? It will be handled by the throwing?

SandGrainOne: In that case it would be better to remove the try catch completely. There are exception handling in the middleware, call chain, and request pipeline. 

However, any unhandled exception will be logged and converted to a 500 internal server error response. Is that what we want here?

src/Storage/Controllers/InstancesController.cs
SandGrainOne: I think the cast will fail if party id is missing and that is the most common case. Further down in the controller we also have more party id logic where we try to convert organization number or national identity number to a party id. Not very common, but too late for the check being done here. 

Application owner will never be given any type or role for an instance owner. I believe its safe to exclude instance owner party id from the authorization request where we know that "access-subject" is an application owner.

src/Storage/Controllers/InstancesController.cs
mikaelrss: Hm. I guess if any of these try-catches actually throw an error, the application haven't been able to determine whether or not the request should have access. I think any sort of 4xx might be confusing for the caller. It would be better if it returned 500 internal server error. 

What do you think?

src/Storage/Controllers/InstancesController.cs
mikaelrss: `DecisionHelper.CreateDecisionRequest` expects a `instanceOwnerPartyId`. Do you have any suggestions as to how to solve this? Will i have to go into Altinn.Common and add a method for creating a `DecisionRequest` without passing `instanceOwnerPartyId`?

SandGrainOne: Looks like we would need to expand AuthorizationService with a new method for this case.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -118,13 +118,28 @@ public async Task<ActionResult<QueryResponse<Instance>>> GetInstances(InstanceQu
...
+                    XacmlJsonResponse response = await _authorizationService.GetDecisionForRequest(request);
+                    
+                    if (!DecisionHelper.ValidatePdpDecision(response?.Response, HttpContext.User))
+                    {
+                        return Forbid();
+                    }
+                }
+                else
+                {
+                    appOwnerRequestingInstances = true;
```
SandGrainOne, line 141: I was thinking that we can allow this to be true for all application owners. Even when not the actual owner of the specific app.

mikaelrss: Alright!

src/Storage/Controllers/InstancesController.cs
mikaelrss: Fixed with new Altinn.Common.PEP v4.1.2

SandGrainOne (approved):

PR #590, opened 2024-12-19
A1/A2: Watermark changes and browser inlining of content

HenningNormann:
## Description
- Give download filename to browser even if content is inline
- Moved watermark generation from html to pdf
   - Write watermark with PdfSharp
   - Added SegoeWpFontResolver to use simple built-in PdfSharp fonts to avoid installing fonts on the alpine image

## Related Issue(s)
- #439 

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Controllers/ContentOnDemandController.cs
```diff
@@ -164,16 +178,25 @@ public async Task<Stream> GetFormdataAsPdf([FromRoute] string org, [FromRoute] s
                     }
                 }
 
-                MemoryStream mergedPdf = new();
-                mergedDoc.Save(mergedPdf);
-                return mergedPdf;
+                pdfStream = new MemoryStream();
+                mergedDoc.Save(pdfStream);
```
Ahmed-Ghanam, line 182: Should we reset the stream position to the beginning before starting the read operation? If not, the read operation will start from the current position (likely the end of the stream), resulting in no data being read.

src/Storage/Controllers/ContentOnDemandController.cs
```diff
@@ -164,16 +178,25 @@ public async Task<Stream> GetFormdataAsPdf([FromRoute] string org, [FromRoute] s
...
+
+            instance.DataValues.TryGetValue("A2ArchRefTs", out string watermark);
+            watermark = (watermark ?? ((DateTime)instance.Created).ToLocalTime().ToString("dd-MM-yyyy HH:mm:ss", CultureInfo.InvariantCulture))
+                + $" AR{instance.DataValues["A2ArchRef"]}";
+
+            using var finalPdfDocument = PdfReader.Open(pdfStream, PdfDocumentOpenMode.Modify);
+            AddWaterMarksAndPageNumber(finalPdfDocument, watermark);
+            MemoryStream finalPdfStream = new();
+            finalPdfDocument.Save(finalPdfStream);
+            return finalPdfStream;
```
Ahmed-Ghanam, line 199: We can reset the stream position to the beginning here as well. This ensures that when the caller reads from the stream, they start from the beginning of the PDF content, avoiding potential issues with empty reads.

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: Read operation: Do you mean the read from the caller of the storage endpoint? I have tested the endpoint and verified that I get a valid pdf that is shown in the browser

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: Ref. previous comment

src/Storage/Controllers/ContentOnDemandController.cs
```diff
@@ -237,13 +260,58 @@ private async Task<string> GetFormdataAsHtmlString(string app, Guid instanceGuid
...
+        }
+
+        private static void AddWaterMarksAndPageNumber(PdfDocument document, string watermark)
+        {
+            for (var idx = 0; idx < document.Pages.Count; idx++)
+            {
+                var page = document.Pages[idx];
+
+                // Get an XGraphics object for drawing beneath the existing content.
+                XGraphics gfx = XGraphics.FromPdfPage(page, XGraphicsPdfPageOptions.Append);
```
Ahmed-Ghanam, line 275: Ensure proper management of resources, particularly the `XGraphics` object. While `XGraphics` does not implement `IDisposable`, it is important to confirm that all associated resources are released appropriately to prevent potential memory leaks or performance issues.
e.g. `using (XGraphics gfx = XGraphics.FromPdfPage(page, XGraphicsPdfPageOptions.Append))`

HenningNormann: Done

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: After finalPdfDocument.Save(finalPdfStream); the Position property of the finalPdfStream is 0 (verified in the debugger).

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: The Position property of finalPdfStream is 0 when returning - verified in the debugger.

PR #598, opened 2025-01-09
Add support for Norwegian locale and time zones

HenningNormann:
## Description
- Add support for Norwegian locale and time zones

## Related Issue(s)
- #439 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,9 @@ WORKDIR /app
 COPY --from=build /app_output .
 COPY --from=build /Storage/Migration ./Migration
 
+# Add support for not only English cultures
+RUN apk add icu-dev icu-libs icu-data-full
+
 # setup the user and group
 # the user will have no password, using shell /bin/false and using the group dotnet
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,6 +18,10 @@ WORKDIR /app
 COPY --from=build /app_output .
 COPY --from=build /Storage/Migration ./Migration
 
+# Add support for not only English cultures
+RUN apk add --no-cache icu-libs icu-data-full tzdata
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false
```
SandGrainOne, line 23: I'm a little worried that changing this might change behavior in unintended ways around in the Storage application.

SandGrainOne: We'll do a before and after comparison of a few API responses.

SandGrainOne (approved):

SandGrainOne:
I've compared the result of a few requests before and after the change and I see no concern to worry. 

I tested:
- Get all applications
- Get all instances for a party (user login)
- Get single instance
- Get specific data blob (Data blob did not have any date fields, but Storage isn't involved in serialization anyway.)

Issue #501, opened 2024-09-20
Support system user in Storage


SandGrainOne:
### Description

This is a list of tasks and changes that needs to be done before we can say that Storage support system users.

### Additional Information

This issue is based on the analysis performed in #471.

Usecase: A system user (representing a user - i.e. sending a party-id) instantiates an instance and populates it with data. E.g. The MVA-report.

#### Authorizaion
- [x] Update to a version of the [Altinn.Common.PEP](https://www.nuget.org/packages/Altinn.Common.PEP#versions-body-tab) package with support for SystemUser authorization (version later than v4.0.0)
- [x] Update all custom Authorization logic to include system user as AccessSubject in Authorization requests.
    The assumption is that we're mostly using the PEP package and the containing DecisionHelper, but there might be exceptions.

#### Metadata updates:
- [x] Search for places where the Intance.CreatedBy property is set. Update the logic to use the organization number of the system owner when caller is a system.
- [x] Search for places where the Intance.LastChangedBy property is set. Update the logic to use the organization number of the system owner when caller is a system.
- [x] Search for places where the DataElement.CreatedBy property is set. Update the logic to use the organization number of the system owner when caller is a system.
- [x] Search for places where the DataElement.LastChangedBy property is set. Update the logic to use the organization number of the system owner when caller is a system.
- [x] Expand the PlatformUser class with properties for: SystemUserId (guid), SystemUserOwner (string) and SystemUserName.
- [x] Search for all usage (assignments) of the PlatformUser class and populate the new properties when data is available.
- [x] Update the logic setting the ProcessHistoryItem.PerformedBy property to support an InstanceEvent created with a system user.
- [x] Update IdentityTelemetryFilter to handle system users

#### Authentication
Not decided yet (optional future improvement):
- [ ] https://github.com/Altinn/altinn-storage/issues/509

### Preparation
Testing this will require multiple steps. 
1. Find a person with the ability to represent an organization. Preferably the "CEO" (daglig leder). Make node of key data we might need. Like names and ids. There are a few Tenor users we've documented on confluence under Guides.
2. We would then need to create a system user. I have not looked into how this could be done yet. The system user would need to be associated with the organization we've identified previously. It doesn't matter what type of system we use. Keep the data like system id and update the test data table on confluence.
3. Find a working app and instantiate an instance or two with the organization as instance owner. (Or find existing instances).
4. Log in with the organization CEO and delegate at a minimum read access to the system user. I have not looked into how this is done at the moment.
[...]

PR #599, opened 2025-01-13
Add support for system user

SandGrainOne:
## Description
This PR is adding basic support for system user as the active/performing party when data is being mutated in Storage.

## Related Issue(s)
- #501 

## Verification
- [x] All tests run green

## Documentation


src/Storage/Services/InstanceEventService.cs
```diff
@@ -7,6 +7,7 @@
 using Altinn.Platform.Storage.Repository;
 
 using Microsoft.AspNetCore.Http;
+using static Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext;
```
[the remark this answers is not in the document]
SandGrainOne: Deleting the statemement.

src/Storage/Filters/IdentityTelemetryFilter.cs
```diff
@@ -42,23 +44,23 @@ public void Process(ITelemetry item)
             DependencyTelemetry dependency = item as DependencyTelemetry;
 
             if (_disableTelemetryForMigration && (
```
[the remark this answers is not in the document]
SandGrainOne: Not touching that monster today.

Ahmed-Ghanam (approved):

src/Storage/Controllers/MessageboxInstancesController.cs
```diff
@@ -311,9 +313,11 @@ public async Task<ActionResult> Delete(Guid instanceGuid, int instanceOwnerParty
                 InstanceOwnerPartyId = instance.InstanceOwner.PartyId,
                 User = new PlatformUser
                 {
-                    UserId = User.GetUserIdAsInt(),
+                    UserId = User.GetUserId(),
                     AuthenticationLevel = User.GetAuthenticationLevel(),
```
martinothamar, line 317: Should we do anything about authentication level for system users? altinn-authentication adds authlevel 3 for Maskinporten token exchange

SandGrainOne: What do you have in mind? Authentication level is already handled as far as I know.

martinothamar: `GetAuthenticationLevel` will return `0`/`null` here since the claim doesn't exist, since systemuser tokens aren't exchanged to Altinn tokens so there is no `authlevel` claim. But technically both a normal service owner token and a system user token is based on the same authentication mechanism (Maskinporten), so it seems strange that the authentication level isn't the same. I'm not sure if the property is used for anything?

PR #603, opened 2025-01-19
Altinn 3 Correspondence bridge to SblBridge

Ceredron:
## Description
Altinn 3 Correspondence has connectivity issues due to running in their own container app environment. Thus they are unable to reach SBL Bridge without making some serious compromises with security configuration. We've seen that Storage already performs a similar operation to SBL Bridge, and that we can easily setup a secure way of calling SBL Bridge through Storage using APIM.

## Related Issue(s)
- [Altinn-Correspondence #603: SBL Bridge via Storage](https://github.com/Altinn/altinn-correspondence/issues/603)

## Verification
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)


src/Storage/Controllers/SblBridgeController.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Endpoint to register Altinn 3 Correspondence recipient in SBL Bridge
+        /// </summary>
+        /// <param name="partyId">The party id that has become an Altinn 3 Correspondence recipient</param>
+        [HttpPost("correspondencerecipient")]
+        [Authorize(Policy = AuthzConstants.POLICY_CORRESPONDENCE_SBLBRIDGE)]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [Produces("application/json")]
+        public async Task<ActionResult> RegisterAltinn3CorrespondenceRecipient(
+            [FromQuery] int partyId)
```
SandGrainOne, line 38: It's a POST request, but the value is a query parameter. Is this on purpose?

Ceredron: Yeah, it is a somewhat idiomatic way to do it when just passing a single parameter. But on further thought, it can be argued that it is sensitive data (and that as a query parameter it will end up in logs) so I will change it to body👍

Ceredron: Fixed.

SandGrainOne (approved):
nit: The url used in the tests still have the query parameter

PR #607, opened 2025-01-21
Altinn 2 code list lookup fixes

HenningNormann:
## Description
- Changed lookup of a2 code lists to be case insensitive
- Avoid infinite loop if a2 code list is not found

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Helpers/A2FormServerXslFunction.cs
```diff
@@ -252,14 +252,17 @@ public XmlDocument GetDOM(string wsInputParams)
                     else
                     {
                         codelist = GetCodeList(wsParams[0], int.Parse(wsParams[1]), int.Parse(wsParams[2]));
-                        foreach (CodeRowBE row in codelist.CodeListRows)
+                        if (codelist.CodeListRows != null)
```
SandGrainOne, line 255: Looks like it is technically possible for `codelist` to be null as well. I didn't dig any deeper than line 368. Would be easier to see for sure with nullable enabled. 
```suggestion
                        if (codelist.CodeListRows != null)
```

SandGrainOne (approved):
Please also consider the suggestions from CodeQL.

PR #625, opened 2025-02-10
Draft: Feature - Deny instance deletion

HauklandJ:
Check DeleteAfterYears attribute in application metadata before permitting soft/hard deletion of instance.

## Description
Checks if an instance is permitted to be deleted before tagging an instance with IsSoftDeleted/SoftDeleted / IsHardDeleted/HardDeleted.

The other option is to check this during cleanup.
The current approach allows us to give feedback to the requester.

## Related Issue(s)
- [#1077](https://github.com/Altinn/app-lib-dotnet/issues/1077)

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Controllers/InstancesController.cs
```diff
@@ -438,6 +438,7 @@ public async Task<ActionResult<Instance>> Post(string appId, [FromBody] Instance
         [ProducesResponseType(StatusCodes.Status204NoContent)]
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
         [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
bjorntore, line 441: Suggest not specifying that we can return 500, as it's an implicit "sh*t hit the fan" response that most, if not all, endpoints can return.

src/Storage/Controllers/InstancesController.cs
```diff
@@ -450,8 +451,32 @@ public async Task<ActionResult<Instance>> Delete(int instanceOwnerPartyId, Guid
...
+                return appInfoError;
+            }
+
             DateTime now = DateTime.UtcNow;
 
+            bool isPreventedFromDeletion = InstanceHelper.IsPreventedFromDeletion(instance, appInfo, now);
+
+            if (isPreventedFromDeletion)
+            {
+                return BadRequest("Instance cannot be deleted yet due to application restrictions.");
```
bjorntore, line 477: Maybe Forbidden instead of BadRequest? Up for discussion.

PR #645, opened 2025-02-24
Language support for watermarks

HenningNormann:
## Description
- Format watermark using user's language

## Related Issue(s)
- #439 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Storage/Controllers/ContentOnDemandController.cs
+++ b/src/Storage/Controllers/ContentOnDemandController.cs
@@ -183,8 +183,18 @@ namespace Altinn.Platform.Storage.Controllers
             MemoryStream pdfStream = new();
             mergedDoc.Save(pdfStream);
 
+            DateTime created;
+            if (instance.DataValues.TryGetValue("A2ArchRefTs", out string a2ArchRefTs))
+            {
+                created = DateTime.Parse(a2ArchRefTs);
+            }
+            else
+            {
+                created = ((DateTime)instance.Created).ToLocalTime();
+            }
+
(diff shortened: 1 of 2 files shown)


src/Storage/Controllers/ContentOnDemandController.cs
```diff
@@ -183,8 +183,18 @@ public async Task<Stream> GetFormdataAsPdf([FromRoute] string org, [FromRoute] s
...
+            if (instance.DataValues.TryGetValue("A2ArchRefTs", out string a2ArchRefTs))
+            {
+                created = DateTime.ParseExact(a2ArchRefTs, "dd-MM-yyyy HH:mm:ss", CultureInfo.InvariantCulture);
+            }
+            else
+            {
+                created = ((DateTime)instance.Created).ToLocalTime();
+            }
+
+            string timestampFormat = language == "en" ? "MM/dd/yyyy hh:mm:ss tt" : "dd.MM.yyyy HH:mm:ss";
```
hggutvik, line 196: You could store the the datetime formats, especially `dd.MM.yyyy HH:mm:ss`, in a const variable. And reference the const in line 189 and here on 196. Could be an advantage to enforce that the same format is used both places

hggutvik (approved):
Approving 🚀 , with one comment that you can consider, about storing datetime format(s) in consts, rather than hard-coding

src/Storage/Controllers/ContentOnDemandController.cs
HenningNormann: All three literals are unique

PR #660, opened 2025-03-17
Log postgres parameters only once

HenningNormann:
## Description
- Log postgres parameters only once
- Avoid logging postgres FirstResponseEvent

## Related Issue(s)
- #[136(team-core-private)](https://github.com/orgs/Altinn/projects/20/views/11?pane=issue&itemId=94396448&issue=Altinn%7Cteam-core-private%7C136)

## Verification
- [ ] All tests run green

## Documentation


src/Storage/Configuration/PostgreSqlSettings.cs
```diff
@@ -18,6 +18,6 @@ public class PostgreSqlSettings
         /// <summary>
         /// Gets or sets a value indicating whether to include parameter values in logging/tracing.
         /// </summary>
-        public bool LogParameters { get; set; } = false;
+        public bool LogParameters { get; set; } = true;
```
hggutvik, line 21: I know that enabling logging is already enabled prior to this PR, but I just wanted to check. Are we sure that sensitive user information isn't leaking into our logs through the parameters?

HenningNormann: I can't imagine that any of the parameters in storage are sensitive and that we would expose more than we already log/trace other ways.

hggutvik (approved):
🚀

PR #676, opened 2025-03-31
Change process next authorization to be more lenient. More actions are accepted for running process next. Depends on task type.

bjorntore:
Change process next authorization to be more lenient. More actions are accepted for running process next. Depends on task type.

## Description
We are in the process of changing the app-lib-code to accept more than one action as a valid process next action. For instance, in a signing task, we allow running process next both with action=sign and with no action, which is equal to action=write.

This is done to support that both these scenarios:
- Singing using process next with action sign.
- Singing using ActionController and then running a process next with no action.

```
POST /action/sign
POST /action/sign
POST /action/sign
PUT /process/next //Fails if number of signatures is less than minCount, or if other validation fails.

//AND
PUT /process/next { action=sign }
```

https://github.com/Altinn/app-lib-dotnet/blob/feature/signing/src/Altinn.App.Core/Internal/Process/ProcessEngineAuthorizer.cs#L34

## Related Issue(s)
Fixes https://github.com/Altinn/app-lib-dotnet/issues/1165

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

- **New Features**
	- Expanded authorization checks to allow multiple valid actions per task type when updating process states.
	- Updated authorization policy to permit signing, writing, and custom task actions for specific user roles in designated process steps.

- **Tests**
	- Added tests to ensure users with appropriate rights can update process states for various task types.
	- Added tests to verify the correct set of actions allowed for each task type.

### DIFF
--- a/src/Storage/Controllers/ProcessController.cs
+++ b/src/Storage/Controllers/ProcessController.cs
@@ -301,7 +301,7 @@ namespace Altinn.Platform.Storage.Controllers
                 "payment" => ["pay", "write"],
                 "confirmation" => ["confirm"],
                 "signing" => ["sign", "write"],
-                _ => [taskType, "write"],
+                _ => [taskType]
             };
         }
     }
--- a/src/Storage/Controllers/ProcessController.cs
+++ b/src/Storage/Controllers/ProcessController.cs
@@ -292,7 +292,7 @@ namespace Altinn.Platform.Storage.Controllers
         /// <summary>
         /// Get all actions that allow process next for the given task type. Meant to be used to authorize the process next when no action is provided.
         /// </summary>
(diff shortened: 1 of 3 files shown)


bjorntore:
@SandGrainOne @TheTechArch I made a suddle change that you might want to look a little closer at. 

TaskId will be included on the decision request every time, if it's not null. I *believe* this won't change any behavior in authorization, other than that they might have enough data to be able to make the decision without having to fetch the instance from storage.

If `processState.CurrentTask.FlowType != "CompleteCurrentMoveToNext"` we seem to use the new process state to authorize. I've tried to keep that as it was.

TheTechArch (approved):
Seems to work based on my reading of the code

sigridbra:
Assume this will be tested by Apps team

Issue #670, opened 2025-03-26
Prevent 500 responsen when client cancels a request


SandGrainOne:
### Description
A consequence of adding better CancellationTokens in Storage we have seen a small increase in the production of 500 responses. This triggers alarms and unecessary noise for the "night guard". We want to implement explicit handling of the System.OperationCanceledException that occur. 

The System.OperationCanceledException exception might be the inner exception of an Npgsql.PostgresException. I'm not sure how to read the stack traces.

It looks like there are 2 cases of this. The new case were we've started to return 500 and an old case where no response is made. The last case is less important, but we should ideally handle both in a way that results in 499.

In these cases we want to return 499 instead.

### Tasks
- [x] Implement graceful handling of the operation canceled exception case.
- [x] Implement handling of any cancellation gracefully by logging as a warning as opposed to exceptions.

### Acceptance criteria
- [x] Storage returns a 499 respons for all cases where a client disconnects.

SandGrainOne:
One more week before production deploy of the latest changes.

sigridbra:
@SandGrainOne This can be unblocked and verified in production now, right?

PR #681, opened 2025-04-08
Change response code to 499 after client cancellation

SandGrainOne:
## Description
I found one case were we're not changing reponse code from 500 to 499 when there has been a cancellation.

## Related Issue(s)
- #670 

## Verification
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

### DIFF
--- a/src/Storage/Controllers/MessageboxInstancesController.cs
+++ b/src/Storage/Controllers/MessageboxInstancesController.cs
@@ -109,7 +109,7 @@ namespace Altinn.Platform.Storage.Controllers
 
                 if (queryResponse?.Exception != null)
                 {
-                    return StatusCode(500, queryResponse.Exception);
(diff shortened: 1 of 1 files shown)


src/Storage/Controllers/MessageboxInstancesController.cs
```diff
@@ -109,7 +109,7 @@ public async Task<ActionResult> SearchMessageBoxInstances([FromBody] MessageBoxQ
 
                 if (queryResponse?.Exception != null)
                 {
-                    return StatusCode(500, queryResponse.Exception);
+                    return StatusCode(cancellationToken.IsCancellationRequested ? 499 : 500, queryResponse.Exception);
```
HenningNormann, line 112: How can you be sure that the cause of the exception is a client side timeout? I think you need to check to cancelllationToken also.

HenningNormann (approved):

PR #723, opened 2025-06-06
Use column confirmed for instance search

HenningNormann:
## Description
- Added filtered index instances_lastchanged_org_not_confirmed_by_org
- Changed search logic to use confirm column when exclude is specified for invoking org
- Changed exclude search definition from string to stringvalues
- Added unit tests

## Related Issue(s)
- https://github.com/Altinn/team-core-private/issues/258

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- **New Features**
  - Added support for filtering instances by confirmation status in queries.
  - Enhanced query parameters to allow excluding confirmed instances by multiple stakeholders.
- **Bug Fixes**
  - Improved filtering logic to ensure accurate retrieval of instances based on confirmation criteria.
- **Tests**
  - Added unit tests to validate filtering behavior related to confirmation status and stakeholder exclusions.


test/UnitTest/TestingRepositories/InstanceTests.cs
```diff
@@ -654,6 +694,106 @@ public async Task Instance_GetInstancesFromQuery_MatchFromPresentationFields_Ok(
...
+            // Arrange
+            Instance newInstance = TestData.Instance_1_1.Clone();
+            newInstance.CompleteConfirmations = new() { new() { StakeholderId = "SKD", ConfirmedOn = DateTime.Now } };
+            await _instanceFixture.InstanceRepo.Create(newInstance, CancellationToken.None);
+
+            InstanceQueryParameters queryParams = new()
+            {
+                Size = 100,
+                ExcludeConfirmedBy = new Microsoft.Extensions.Primitives.StringValues(["SKD"]),
+                Org = "SKD"
```
SandGrainOne, line 762: Should the org parameter be TTD?

The primary purpose of the Org parameter is to include instances for apps owned by the org. In this test the instance belong to TTD. The result should be 0 instances even without the ExcludeConfirmedBy parameter.

src/Storage/Models/InstanceQueryParameters.cs
```diff
@@ -348,26 +358,38 @@ private static void AddDateParamIfNotNull(Dictionary<string, object> postgresPar
...
+            {
+                return Confirmed.Value;
+            }
+
+            return ExcludeConfirmedBy.HasValue && ExcludeConfirmedBy.Contains(Org) ? false : null;
+        }
+
         /// <summary>
-        /// Retrieves an array of exclude confirmed by values from the query values.
+        /// Retrieves an array of exclude confirmed by values that differ from the invoking org.
```
SandGrainOne, line 372: "Invoking org" can be a bit misleading. Remember that these parameters can be used by end user systems as well. The Org parameter is simply an instance filter like any other.

test/UnitTest/TestingRepositories/InstanceTests.cs
HenningNormann: Fixed

PR #782, opened 2025-08-31
Add ASB mapper for deduplication support

elsand:
This adds a Azure Service Bus message mapper that deterministically sets `MessageId` based on the instance id and a 5 second time bucket. Additionally, it schedules the message to be delivered at the end of the bucket period, to ensure that any consumers will fetch the most recent version of the instance within that bucket.

This will enable the ASB duplicate detection mechanism to silently drop redundant messages for updates for any given instance, at the cost of a slight delay (up to 5 seconds) for consumers (dialogporten-adapter) to see the message. The delay will ensure that no user updates are missed, while the duplication will reduce the number of messages being marked as active in the queue, causing a reduction of load on both the storage and Dialogporten APIs.

NOTE! Duplicate detection can only be set upon creation of a ASB queues, so for exising queues they will have to be deleted first and recreated with something like

```
az servicebus queue delete --resource-group my-resource-group  --namespace-name my-asb-namespace --name "altinn.dialogportenadapter.webapi"
az servicebus queue create --resource-group my-resource-group  --namespace-name my-asb-namespace --name "altinn.dialogportenadapter.webapi" --enable-duplicate-detection true  --duplicate-detection-history-time-window PT20S
```

## Summary by CodeRabbit

- New Features
  - Coalesces rapid update bursts into a single scheduled sync within a short time window.

- Bug Fixes
  - Reduces duplicate sync messages and improves delivery reliability to avoid missed updates.

- Chores
  - Enables queue duplicate-detection and a durable outbox for outgoing messages to strengthen message handling.

- Tests
  - Adds tests verifying time-bucketed deduplication and scheduling behavior.


src/Storage/Helpers/DebounceHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// </summary>
+    public static (Guid Id, DateTimeOffset BucketEndUtc) TimeBucketId(string resourceId, TimeSpan bucket, DateTimeOffset nowUtc)
+    {
+        var utcTicks = nowUtc.UtcTicks;
+        var bucketStartUtcTicks = utcTicks - (utcTicks % bucket.Ticks);
+        var bucketStartUtc = new DateTimeOffset(bucketStartUtcTicks, TimeSpan.Zero);
+        var bucketEndUtc = bucketStartUtc + bucket;
+
+        var key = $"{resourceId}|{bucketStartUtc:O}|{bucket.Ticks}";
+        using var sha = SHA256.Create();
```
MagnusSandgren, line 24: Denne kan leve under hele applikasjonens levetid som en private static field på DebounceHalper

src/Storage/Program.cs
```diff
@@ -360,11 +360,35 @@ void ConfigureWolverine(IServiceCollection services, IConfiguration config)
...
             opts.UseAzureServiceBus(wolverineSettings.ServiceBusConnectionString)
 
+                // Use custom mapper to set deduplication id on outgoing messages
+                .ConfigureSenders(s =>
+                {
+                    s.CustomizeOutgoingMessagesOfType<SyncInstanceToDialogportenCommand>((envelope, cmd) =>
+                    {
+                        // Set a deterministic MessageId for SyncInstanceToDialogportenCommand messages, and
+                        // deliver them at the end of a time bucket.
+                        var (id, bucketEndUtc) = DebounceHelper.TimeBucketId(cmd.InstanceId, 5.Seconds(), DateTimeOffset.UtcNow);
```
MagnusSandgren, line 370: Smart! ASB har 20 sec som minste vindu, men du lager en id basert på 5 sec bucket!

src/Storage/Helpers/DebounceHelper.cs
elsand: Tror ikke det er thread-safe? Men uansett, det er bedre å bruke den statiske `SHA256.HashData` for å slippe `IDisposable`

PR #868, opened 2025-11-26
Run CSharpier on the solution

ivarne:
Now that I'm eating cake for Altinn-storage, I think we should align the style with the other studio repos

1. Remove stylecop (it seems to be incompatible with CSharpier)
2. Convert all files to file scoped namespaces
3. Add the tooling for csharpier (A `Directory.Build.props` file in the root of the project that runs CSharpier on build)
4. Run CSharpier on all files in the solution

Remaining tasks
1. Uncomment "EnableNETAnalyzers"
2. Set TreatWarningsAsErrors to True when running in CI

## Summary by CodeRabbit

* **New Features**
  * Many new and expanded storage APIs: applications, instances, data upload/download, process actions, events, messagebox, signing, migration, on‑demand HTML/PDF and studio endpoints.

* **Chores**
  * Added project tooling and formatter configuration; consolidated namespaces and flattened nested types for a cleaner codebase.

* **Documentation**
  * Improved API docs and XML comments across controllers and helpers for clearer behavior and responses.

### DIFF
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -2,6 +2,8 @@ name: Verify dotnet format
 
 on:
   pull_request:
+permissions:
+  contents: read
 
 jobs:
   verify-no-changes:
--- a/src/Storage/Authorization/AuthorizationService.cs
+++ b/src/Storage/Authorization/AuthorizationService.cs
@@ -47,7 +47,7 @@ public class AuthorizationService(
     private const string ActionId = "a";
     private const string ResourceId = "r";
 
-    /// <inheritdoc/>>
+    /// <inheritdoc />
     public async Task<List<MessageBoxInstance>> AuthorizeMesseageBoxInstances(
         List<Instance> instances,
         bool keyAccessMode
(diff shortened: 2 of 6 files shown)


.csharpierignore
```diff
@@ -0,0 +1,7 @@
+bin/
+obj/
+.git/
+*.xml
+*.csproj
+*.props
+*.targets
```
HauklandJ: We should probably ignore migrations: 

**/Migration/

Directory.Build.props
```diff
@@ -0,0 +1,22 @@
...
+  </PropertyGroup> -->
+
+  <!-- <PropertyGroup>
+    <LangVersion>latest</LangVersion>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <AnalysisMode>Minimum</AnalysisMode>
+    <Features>strict</Features>
+    <NuGetAudit>true</NuGetAudit>
+    <NuGetAuditMode>all</NuGetAuditMode>
+  </PropertyGroup> -->
```
HauklandJ: Why is the commented out code left in?

HauklandJ: I read the description now, resolving this comment

.csharpierignore
ivarne: Why? Is there any damage to migration files?

ivarne: I don't see any changes in folders named `Migration`, so i don't see why this would be an issue.

PR #869, opened 2025-11-26
Replace legacy k6 action in workflows

hggutvik:
### Background
`k6-action` [was abandoned Jul 16, 2024](https://github.com/grafana/k6-action?tab=readme-ov-file) and has not been maintained since. It's in use in use-case-* and regression-test-* workflows.

The k6 workflows now follow the new two-step pattern where:

grafana/setup-k6-action sets up the k6 environment once per job
grafana/run-k6-action runs the individual k6 test files with the specified parameters

### Changes Made:
- Formats the YAML of the relevant files (using [the Red Hat YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml))
- Replaces grafana/k6-action ([abandoned Jul 16th, 2024](https://github.com/grafana/k6-action?tab=readme-ov-file)) with `grafana/setup-k6-action` and `grafana/run-k6-action`
  - `setup-k6-action` is run initially, in a single step after git checkout
  - `run-k6-action` is run for each test
  - breaking changes:
    (i) action input `filename` from the old k6-action must be replaced with `path`
    (ii) handling of cases where the js script **at module level** (i.e. outside of functions) depends on defined values for any environment variable
      - Reason: a new pre-run k6 validation process (`k6 inspect`) loads and validates the script **_before_** the environment variables from the GitHub workflow are available to the k6 script via `__ENV`. That means that the k6 script now either (a) must tolerate `undefined` for that env var OR (b) a value for the environment variable has to be passed as a flag via the input `inspect-flags`.
      Note that this only applies to environment variables that the script evaluates at module level, i.e. code that runs when k6 loads the module. `k6 inspect` does not run all the script's functions - it only runs functions if they are run at module level, [e.g. like the `open` function is run here](https://github.com/Altinn/altinn-storage/blob/main/test/k6/src/tests/data.js#L29C19-L29C56). This means that  env var access [like this](https://github.com/Altinn/altinn-storage/blob/main/test/k6/src/tests/applications.js#L24) does not need remediation in this PR, as the evaluation of that env var does not occur in the inspect phase (only later when the test script is actually run).
      
      For Storage, this only applies to (i) the `app` env (only for the k6 test script `data.js`), and (ii) the variable holding the Altinn environment ("AT22", "TT02", etc.). The latter variable was previously declared as `env`, renamed in this PR to `altinn_env` for a clearer separation from [the `env` keyword](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#env). 

- Reorganizes passing of env variables from the GitHub workflow job to the k6 scripts
   Moves common k6 flags across scripts to a shared job `env` block. This still makes them accessible to the k6 js scripts via the `__ENV` object, without having to repeatedly pass them as separate flags in all k6 actions
   ➕ improves readability and simplifies maintenance by reducing the duplication of identical flags passed to many k6 scripts.
   ➖ Sharing of variables within a job turns to oversharing if (a) the job includes third-party GitHub actions and (b) the variables include sensitive values. Thus creating the need for the change below 👇 
- Isolates the Slack notification step in a separate job
  - ➕ allows us to [restrict what secrets and env vars should be available](https://github.com/Altinn/altinn-storage/pull/869/changes#diff-4842ade78b195eb6dc8fe726ac74260c9943137dc35907e87d2411095db91088R81) to the third-party GitHub action for posting to Slack
  - ➕ allows use of a [reusable workflow for posting to Slack](https://github.com/Altinn/altinn-storage/pull/869/changes#diff-2f6ccf78e50e4905bb7a57393d310e0ad2bd8e86c71be65e969c41fe14607c54), which simplifies the calling workflows (e.g. [as in use-case-TT02](https://github.com/Altinn/altinn-storage/pull/869/changes#diff-46850e1727caf8bcefe98ce1f2f224340fcd85201e024cb2d50c600e3268a22cR69))
  - ➖ the use-case-ATX workflow gets more complex than it was previously. Moving the Slack action to a separate job means that it will run **after** the whole matrix of [the `usecase-test` job](https://github.com/Altinn/altinn-storage/pull/869/changes#diff-46850e1727caf8bcefe98ce1f2f224340fcd85201e024cb2d50c600e3268a22cR69) has completed. That is, the k6 tests have to run for AT22, AT23, and AT24 first, _and then_ the Slack notification job can run. To enable the Slack notification job to know _which_, if any, matrix environments got failed tests, the `test` job now has to communicate this via job outputs ([link to the relevant code change](https://github.com/Altinn/altinn-storage/pull/869/changes#diff-ce7b1da48e1d748ff3e9e89ef703b46fc5d183531ccc685753a9dcf875a42576R13)).
I can't see a way to integrate the new reusable Slack workflow in this workflow's slack-notify job because of the matrix logic

## Related Issue(s)
https://github.com/Altinn/team-core-private/issues/266 (private aggregate issue for replacing the `k6-action` in all of our repos)

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **Chores**
  * Normalized YAML formatting/indentation across CI workflows and ensured trailing newlines.

* **Refactor**
  * Migrated test workflows to a unified k6 setup-and-run pattern with shared environment blocks and consolidated failure handling into a separate notification job.
  * Aligned test invocation to path-based execution and standardized step structure.

* **New Features**
  * Added a reusable Slack warning workflow for failure notifications.
  * Test runtime now reads a consolidated environment key, affecting which test data/env is selected.

### DIFF
--- a/.github/workflows/regression-test-PROD.yml
+++ b/.github/workflows/regression-test-PROD.yml
@@ -37,7 +37,6 @@ jobs:
           path: test/k6/src/tests/instances.js
           flags: -e encodedJwk=${{ secrets.MP_ENCODEDJWK }} -e mpClientId=${{ secrets.MP_CLIENT_ID }} -e mpKid=${{ secrets.MP_KID }} -e orgNumber=${{ secrets.ORG_NUMBER }} -e orgPartyId=${{ secrets.ORG_PARTY_ID }} -e pid=${{ secrets.REGRESSION_PERSON_NUMBER }}
           inspect-flags: -e altinn_env=PROD # Supplies the value for pre-run validation, when env vars / flags are not yet defined
-
       - name: Run regression tests for endpoint 'instances/sbl'
         if: always()
         uses: grafana/run-k6-action@a15e2072ede004e8d46141e33d7f7dad8ad08d9d # v1.3.1


Konrad-Simso (changes requested):
Very nice work, i've put in some questions on whitespace formatting out of curiosity.
Also, I think there is still some testing code left in the `use-case-ATX.yml` file.

Last question, i see that there are different values from secrets used for the variables `userName` and `userPassword` for TT02 and PROD, is this meant to be different, and if so, could have have env related names? (This might be difficult or out of scope, not required)

Konrad-Simso (approved):

PR #879, opened 2025-11-28
Pass cancellation token CA2016

ivarne:
When enabeling NETAnalyzers we get warnings when a method has a cancellation token but does not pass it forward to other methods that has cancellationTokenParameter

It is not always correct to pass a cancellationToken (eg, after saving part 1 of something that should have been protected by a transaction), but here I (hopefully) only fix the obvious cases.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced data operation handling across storage controllers with improved cancellation support for better system responsiveness during long-running operations.

### DIFF
--- a/src/Storage/Controllers/CleanupController.cs
+++ b/src/Storage/Controllers/CleanupController.cs
@@ -171,7 +171,11 @@ public class CleanupController(
                         false,
                         cancellationToken
                     );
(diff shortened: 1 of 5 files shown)


Konrad-Simso (approved):
✅ LGTM, nice work.

PR #965, opened 2026-04-23
feat: enable nullable by default

Konrad-Simso:
## Description
Enable nullable by default on all solutions.

## Related Issue(s)
- https://github.com/Altinn/altinn-studio/issues/18423

## Verification
- [x] All tests run green

### DIFF
--- a/test/UnitTest/TestingServices/SigningServiceTest.cs
+++ b/test/UnitTest/TestingServices/SigningServiceTest.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+﻿#nullable disable
 
 using System;
 using System.Collections.Generic;
@@ -304,7 +304,7 @@ public class SigningServiceTest
             Signee = signee,
         };
 
-        string? performedBy = !string.IsNullOrWhiteSpace(signee.UserId)
+        string performedBy = !string.IsNullOrWhiteSpace(signee.UserId)
             ? signee.UserId
             : signee.OrganisationNumber;
         (bool created, ServiceError serviceError) = await service.CreateSignDocument(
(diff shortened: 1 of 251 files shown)


test/UnitTest/TestingServices/SigningServiceTest.cs
```diff
@@ -1,4 +1,4 @@
-﻿#nullable enable
+﻿#nullable disable
```
Konrad-Simso, line 1: This file became a problem because of some mocked service methods. 
Turning off nullable for this file until a broader  pass for nullability of types is done.

HauklandJ (approved):
Looks good, nicely scoped PR. I give an approval assuming all files without any nullable directive now have nullable disabled.

Konrad-Simso:
> I give an approval assuming all files without any nullable directive now have nullable disabled.

I do have a question about that, should i add `#nullable enable` on all files, even if it doesn't need to have it? Parts of the solution already handle nullability correctly.

HauklandJ:
> > I give an approval assuming all files without any nullable directive now have nullable disabled.
> 
> I do have a question about that, should i add #nullable enable on all files, even if it doesn't need to have it? Parts of the solution already handle nullability correctly.

Here is the algorithm:

```
while(files_not_checked > 0):
    var file = files_not_checked.pop()
    if (nullable_enabled(file)):
        remove_nullable_directive(file)
    if (no_nullable_directive_exsts(file)):  // neither enable nor disabled
        add_disable_nullable(file)
```

Konrad-Simso:
Then i'll do another pass over all files

PR #977, opened 2026-05-13
Handle `GeneratedFromTask` cleanup in `PutInstanceAndEvents`

danielskovli:
## Summary

- Adds `IProcessDataCleanupService` / `ProcessDataCleanupService` and invokes it from `ProcessController.PutInstanceAndEvents` whenever the incoming `ProcessState.CurrentTask.ElementId` is non-empty.
- For the target task, finds all data elements on the instance whose `References` contain `{Relation = GeneratedFrom, ValueType = Task, Value = <targetTaskId>}` and deletes them via the existing `IDataService.DeleteImmediately` path (blob + row + `Deleted` event).
- Cleanup runs **after** authorization and **before** the instance/event Postgres transaction. Per-element failures are logged and skipped so cleanup is best-effort and never blocks the process advance.
- No new endpoints, no public API surface changes — purely an internal addition to existing controller behavior.

## Why

Tracks three issues that turned out to be different surface effects of the same problem (cleanup running in app-lib against the in-memory `Instance` rather than the persisted state):
- Altinn/altinn-studio#18504 — symptom: cleanup fails when user lacks write permission during `process/next`.
- Altinn/altinn-studio#18465 — proposal: handle cleanup server-side as part of the same Storage transaction.
- Altinn/app-lib-dotnet#1735 — symptom: legacy PDF generation leaves a stale PDF when a data task loops back to itself before exiting to an EndEvent.

The app-lib used to perform this cleanup itself in `ProcessTaskInitializer` and `SubformPdfServiceTask` by calling the data-delete endpoint, which is guarded by `POLICY_INSTANCE_WRITE` + `dataTypeDefinition.CanWrite()`. Users authorized to advance the process but not to write the affected data type hit a permission failure mid-cleanup. Moving the logic into `PutInstanceAndEvents` lets Storage perform the deletes under the process-advance authorization context, eliminating that failure mode — and as a side effect resolves #1735 because Storage operates on the persisted instance loaded from DB, which does see the legacy PDF.

## Scope

This is a lift-and-shift of the existing app-lib filter (`Relation = GeneratedFrom && ValueType = Task && Value = targetTaskId`). It deletes the same set of data elements the app-lib used to delete — no broader. Untagged data elements (form data, signing's signee-state JSON, `AddBinaryDataElement` callers) remain untagged and are unaffected by this change; closing those gaps is a separate workstream tracked by the TODO at `SigningService.cs:74` in app-lib.

## Companion PR

- https://github.com/Altinn/app-lib-dotnet/pull/1750. Removes the now-redundant `ProcessTaskCleaner` and its two call sites. That PR is blocked on this one shipping first.

## Rollout

- **Deploy Storage first**, then merge and release the app-lib PR.
- Window between Storage deploy and app-lib release: old app-lib still runs `ProcessTaskCleaner` before its `process/next` PUT, so the same elements are deleted twice. Both legs are idempotent:
  - `storage.deletedataelement_v2` uses `DELETE ... WHERE alternateid = ...` — no-op on missing row.
  - `BlobRepository.DeleteBlob` wraps `DeleteIfExistsAsync` — explicitly idempotent.
  - One caveat: `DataService.DeleteImmediately` unconditionally dispatches a `Deleted` instance event afterward, so the rollout window will emit a duplicate event per element. Audit-trail noise only.
- Apps still on the unfixed app-lib version will continue to hit the original auth failure until they upgrade — acceptable per scope.

## Design notes

- **Cleanup placement** — after auth check, before `_instanceAndEventsRepository.Update`. If the subsequent PG tx fails, the cleanup is naturally idempotent on retry (deleted elements are no longer in `instance.Data`).
- **Atomicity** — blob deletes can't enroll in a Postgres transaction, so cleanup is intentionally outside the PG tx. The existing `DataService.DeleteImmediately` ordering (blob → row → event) is preserved.
- **Terminal transitions** — cleanup is skipped when `CurrentTask` is null (process end / abandon), so signed PDFs, receipts, and other archival artifacts are preserved. This is what lets the final-transition PDF in #1735 survive while loop-iteration PDFs are reaped.
- **Filter** — uses all three fields (`Relation`, `ValueType`, `Value`). Matches the issue #18465 proposal and the data-creation pattern in `DataElementHelper.cs`.

## Test plan

- [x] `dotnet build Altinn.Platform.Storage.slnx` — clean (0 warnings, 0 errors)
- [x] New `ProcessDataCleanupServiceTests` (7 tests) — null/empty data, no-match filters, three-field match + mutation, mid-loop exception continues, missing Application fallback, cancellation
- [x] Extended `ProcessControllerTest` (4 new tests) — cleanup invoked with destination task id, terminal transition skips cleanup, unauthorized request skips cleanup, cleanup-service exception bubbles up
- [x] Existing `ProcessControllerTest` suite — all 47 still pass (43 prior + 4 new)
- [ ] End-to-end verification in `Altinn.App.Integration.Tests` alongside app-lib PR #1750: re-entering a task with stale `generatedFromTask` PDFs removes them; subform-PDF service task generates new PDFs that survive the subsequent exit-event PUT; #1735 loop-to-EndEvent scenario leaves exactly one PDF.

## Summary by CodeRabbit

* **New Features**
  * Best-effort automatic removal of data generated by previous tasks when advancing into a new task (skips terminal transitions and unauthorized requests).

* **Behavior**
  * If the cleanup operation throws, the update request returns a server error (500).

* **Tests**
  * Comprehensive tests covering cleanup success, partial failures, cancellation, authorization short-circuits, and application-lookup failure.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/altinn-storage/pull/977?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/Storage/Services/IProcessDataCleanupService.cs
```diff
@@ -0,0 +1,35 @@
+#nullable disable
```
ivarne: ??

src/Storage/Services/ProcessDataCleanupService.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                return application.StorageAccountNumber;
+            }
+
+            _logger.LogWarning(
+                "Could not resolve Application for {AppId}; proceeding with default storage account when deleting stale data elements",
+                instance.AppId
+            );
+        }
+        catch (Exception ex)
```
ivarne: "Gotta catch them all!"

Tror vel egentlig ingenting virker med appen om dette feiler, så kanskje litt defensivt å fange eventuelle exceptions her.

ivarne (approved):
Ser bra ut!

ivarne (approved):

PR #982, opened 2026-05-18
chore: update nuget workflow

Konrad-Simso:
## Description
Update nuget auth to use trusted publishing.


Konrad-Simso:
Closing due to TP tying a user to the token generation.

PR #1022, opened 2026-06-26
feat: idempotent data element creation via idempotencyKey

danielskovli:
## Summary

Adds an optional `idempotencyKey` query parameter to `POST storage/api/v1/instances/{partyId}/{instanceGuid}/data`. When supplied:

- On create, Storage stamps the key into the data element's existing `Metadata` under the reserved key `altinn-idempotency-key` (`DataElementHelper.IdempotencyKeyMetadataName`).
- A repeated request with the **same key for the same instance** returns the already-created element (before reading the body / writing a blob) instead of inserting a duplicate.

Omitting the parameter preserves the previous behaviour byte-for-byte.

## Why

This is the Storage half of a fix for a **P1** in the Altinn.App workflow-engine integration ([altinn-studio#18735](https://github.com/Altinn/altinn-studio/pull/18735)). The workflow engine delivers callbacks **at-least-once**: a callback can persist a data element to Storage but fail to be recorded by the engine (lost response, etc.), so the engine retries the same command. Without an idempotency signal in Storage, the retry inserts a duplicate form-data element.

The key is a durable, retry-stable correlation token. Storage is the only place the signal can live across a retry, so the dedup must happen here.

## Design notes / constraints

- **No schema change.** The marker reuses the existing `Metadata` (`List<KeyValueEntry>`) column; data element IDs are untouched.
- **Backwards compatible & non-breaking.** New optional query param; existing callers are unaffected.
- Callbacks for an instance are processed serially, so there is no concurrent-insert race for a given key.

## Tests

`DataControllerUnitTests`:
- `CreateAndUploadData_WithIdempotencyKey_StampsMarkerOnNewElement`
- `CreateAndUploadData_WithIdempotencyKey_ReturnsExistingElementOnReplay`

## Related PR — must ship together

App side: **Altinn/altinn-studio#19367** (a discrete fix stacked on the workflow-engine integration PR #18735). The App sends the `idempotencyKey`; this PR makes Storage honor it. The App change is safe to deploy before this (the key is simply ignored until Storage honors it), but the dedup only takes effect once both are deployed.

## Summary by CodeRabbit

* **New Features**
  * Added support for idempotent uploads using an optional key, helping prevent duplicate data elements on repeated requests.
  * Existing uploaded data can now be returned directly when the same key is reused.

* **Bug Fixes**
  * Preserved the idempotency marker during metadata updates so it is not lost when updating existing elements.
  * Ensured returned data includes the correct platform links and instance information.

### DIFF
--- a/src/Storage/Controllers/DataController.cs
+++ b/src/Storage/Controllers/DataController.cs
@@ -495,6 +495,9 @@ public class DataController : ControllerBase
             );
             if (existing != null)
             {
+                // Guard against a persisted element with an unset InstanceGuid so the self link / Location is built
+                // from a valid guid rather than producing a malformed URL.
+                existing.InstanceGuid ??= instanceGuid.ToString();
                 existing.SetPlatformSelfLinks(_storageBaseAndHost, instanceOwnerPartyId);
                 return Created(existing.SelfLinks.Platform, existing);
             }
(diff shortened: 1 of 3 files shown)


danielskovli:
Superseded, at least for now, by https://github.com/Altinn/altinn-storage/pull/1049

PR #1047, opened 2026-07-16
fix: reuse single connection pool in unit tests

Konrad-Simso:
## Description
Update `ServiceUtil.cs`  and `ServiceCollectionExtensions.cs` to use common `NpgsqlDataSource`.
When running the tests it was causing errors due to opening too many connection pools at the same time.

## Verification
- [x] All tests run green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL repository DI registration by requiring a single, preconfigured shared database data source (dynamic JSON enabled) for more consistent setup.
* **Tests**
  * Updated outbox repository unit tests to reuse one repository instance and a single opened database connection per test.
  * Improved unit-test reliability by reusing a cached shared DI service provider and a shared database data source.
* **Refactor**
  * Simplified unit-test service provisioning by removing per-call environment/config setup and updating the test helper API.

### DIFF
--- a/test/UnitTest/TestingRepositories/OutboxTests.cs
+++ b/test/UnitTest/TestingRepositories/OutboxTests.cs
@@ -10,7 +10,6 @@ using Altinn.Platform.Storage.Messages;
 using Altinn.Platform.Storage.Repository;
 using Altinn.Platform.Storage.UnitTest.Extensions;
 using Altinn.Platform.Storage.UnitTest.Utils;
-using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -207,12 +206,12 @@ public class OutboxTests
 
         var config = builder.Build();
 
-        WebApplication.CreateBuilder().Build().SetUpPostgreSql(true, config);
-
         IServiceCollection services = new ServiceCollection();
 
         services.AddLogging();
-        services.AddPostgresRepositories(config);
+
+        services.AddSingleton(ServiceUtil.GetSharedDataSource());
+        services.AddRepositoryImplementations();
         services.AddMemoryCache();
 
         Mock<IHttpContextAccessor> httpContextAccessor = new Mock<IHttpContextAccessor>(
--- a/test/UnitTest/TestingRepositories/OutboxTests.cs
+++ b/test/UnitTest/TestingRepositories/OutboxTests.cs
@@ -19,6 +19,7 @@ using Xunit;
 
 namespace Altinn.Platform.Storage.UnitTest.TestingRepositories;
 
+[Collection("StoragePostgreSQL")]
 public class OutboxTests
 {
     public OutboxTests()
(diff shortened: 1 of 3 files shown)


test/UnitTest/Utils/ServiceUtil.cs
```diff
@@ -53,8 +33,61 @@ public static List<object> GetServices(
...
+            services.AddPostgresRepositories(config);
+            services.AddMemoryCache();
+
+            services.Configure<GeneralSettings>(config.GetSection("GeneralSettings"));
+
+            _serviceProvider = services.BuildServiceProvider();
+        }
+
+        return _serviceProvider;
+    }
```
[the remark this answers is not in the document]
Konrad-Simso: Have a look at the last 2 commits pushed after this comment. I've removed the `envVariables` due to the pattern not being used at all in the test suite.

test/UnitTest/Utils/ServiceUtil.cs
```diff
@@ -7,43 +7,23 @@
...
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Npgsql;
 
 namespace Altinn.Platform.Storage.UnitTest.Utils;
 
 public static class ServiceUtil
 {
+    private static readonly object _lock = new();
+    private static ServiceProvider _serviceProvider;
```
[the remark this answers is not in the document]
Konrad-Simso: Pattern has been changed, noe implements the provider with `Lazy<T>`, that should fix this concern at the same time i assume?

Konrad-Simso:
@coderabbitai full review

test/UnitTest/TestingRepositories/OutboxTests.cs
```diff
@@ -190,9 +190,7 @@ private static IOutboxRepository GetRepo(
 
     private static NpgsqlConnection GetConnection()
```
olavsorl, line 205: Claude:

> Pre-existing connection leak in OutboxTests.GetConnection() (not introduced here, but now more relevant): GetSharedDataSource().OpenConnection() returns an open NpgsqlConnection that's passed into Insert(cmd, existingConnection) — and src/Storage/Repository/PgOutboxRepository.cs:69 never disposes it. Each opened connection is never returned to the pool. Previously each leak sat in its own throwaway pool; now they all draw from the single shared pool. The count is small (~10 across the class vs. Npgsql's default max pool size of 100), so it won't exhaust anything — but it's slightly ironic given the PR's goal. Worth a follow-up to wrap these in await using or have the test dispose them.

test/UnitTest/Utils/ServiceUtil.cs
```diff
@@ -7,29 +7,48 @@
...
 
 public static class ServiceUtil
 {
-    public static List<object> GetServices(
-        List<Type> interfaceTypes,
-        Dictionary<string, string> envVariables = null
-    )
+    private static readonly Lazy<ServiceProvider> _serviceProvider = new(BuildServiceProvider);
+
+    public static List<object> GetServices(List<Type> interfaceTypes)
```
olavsorl, line 18: Claude:
> OutboxTests.GetServices still builds a new ServiceProvider per GetRepo() call, and several tests call GetRepo() multiple times. No pool is created (the data source is shared now), so it's just minor allocation churn — not blocking, but resolving the repo once would be tidier.

olavsorl (approved):
LGTM! Claude hadde et par kommentarer thou

test/UnitTest/Utils/ServiceUtil.cs
Konrad-Simso: Updated tests to call `GetRepo` once per test.

test/UnitTest/TestingRepositories/OutboxTests.cs
Konrad-Simso: Updated tests to create one connection per test.

PR #1088, opened 2026-09-03
fix: comment out alter table commands

Konrad-Simso:
## Description
Comment out alter table requests so they can be done safely before migration is applied. Then later uncomment the lines of code for tests.

Tests will fail on this PR. Its meant as a stand alone for deployment.

## Verification
- [ ] All tests run green


Konrad-Simso:
Closing since i found another approach in #1089

PR #1100, opened 2026-09-14
fix: instance mutation endpoint - delete status

Konrad-Simso:
## Description
V8 apps require some changes on instances to be allowed after an instance has been deleted. 
An example is if `autoDeleteOnProcessEnd` is set to true, delete status is set before clean up is done, resulting in data elements not being deleted and instance events not being created.

## Verification
- [x] All tests run green

## Summary by CodeRabbit

* **Bug Fixes**
  * Data-element creations and updates continue to be rejected for hard-deleted instances.
  * Deletes, instance updates, and events can now be applied to hard-deleted instances, including recording process completion and events.

* **Database Updates**
  * Added the database capabilities required for instance and data-element mutations, locking, event handling, blob management, queries, and migration state management.

* **Tests**
  * Added coverage verifying mutation behavior and process/event recording for hard-deleted instances.


vxkc (approved):

Konrad-Simso:
Updated comment, wanted less lines for the general comment in the sql function.

PR #1102, opened 2026-09-15
feat: add api key protected endpoint for deleting a single data element

olavsorl:
## Description

Adds an endpoint for deleting a single data element, its blob and recording a delete event, callable from inside the cluster with curl and no Altinn token:

```
DELETE storage/api/v1/cleanup/dataelement/{instanceOwnerPartyId}/{instanceGuid}/{dataGuid}
```

It is guarded by a shared secret in the `X-API-Key` header instead of being left unauthenticated. The full instance/data element triple is required, so the endpoint cannot be used for bulk deletion, and every call is logged with the deleted element and the caller IP.

The API key validation already existed for the metrics endpoints, so it is extracted into an `ApiKeyAuthorizationFilter` base class shared by `MetricsApiKeyFilter` and the new `CleanupApiKeyFilter`. Each filter passes its own `ILogger<T>` to the base, so log categories are unchanged.

While wiring this up, `DataService.DeleteImmediately` needed two changes:

- It rebuilt the blob path with `DataElementHelper.DataFileName(...)` instead of using the data element's stored `BlobStoragePath`. Since #1075 introduced versioned blob paths (`{appId}/{instanceGuid}/data-elements/{version}`), the delete targeted a path that does not exist for such elements and silently orphaned the blob. It now uses the stored path and logs when no blob was found. This also fixes the same latent problem for `ProcessDataCleanupService` and `SigningService`.
- `DispatchEvent(type, instance, dataElement)` throws when the request carries no user claims, so an unauthenticated caller would have deleted the blob and row and then failed before writing the event. Both `DeleteImmediately` and that `DispatchEvent` overload now accept an optional actor and additional info; when no actor is given the existing claims-based resolution is used unchanged.

## Deployment note

`StorageCleanupApiKey` must be provisioned as a Key Vault secret before the endpoint works. Until it exists the filter fails closed with `503`, so the endpoint is inert rather than open.

## Verification

- `dotnet build` clean, 0 warnings
- `dotnet csharpier check .` clean
- New tests: 6 cases for the shared API key filter, 4 for `DeleteImmediately` (versioned blob path, event with and without an explicit actor, missing blob)
- The DB-backed tests could not be run locally (no Postgres with the expected credentials). `MetricsControllerTests` exercises the refactored filter end to end and should be confirmed green in CI. There is no controller-level test for the new endpoint, since `CleanupController` has no existing test coverage.

## Summary by CodeRabbit

- **New Features**
  - Added an API-key-protected endpoint for immediately deleting individual data elements during cleanup.
  - Cleanup and metrics endpoints now validate API keys consistently, including clear responses for missing, invalid, or unavailable configuration.
  - Cleanup deletions record the initiating system user and the reason for the deletion.

- **Bug Fixes**
  - Immediate deletion now targets the data element’s stored blob location, while still completing successfully when the blob is already missing.
  - Improved client IP detection for authorization logging.


olavsorl:
Superseded by #1103 (shared API key filter) and #1104 (the endpoint).

Main has since restructured data element deletion — `IDataService.DeleteImmediately` no longer exists, and deletes now go through `IInstanceMutationRepository.Apply` so the metadata delete and the instance event are applied as one transaction. The blob-path fix in this PR is obsolete with that change, so rather than resolve a merge where most of the resolution was discarded work, the feature was rebuilt on the new flow.

PR #1104, opened 2026-09-15
feat: add api key protected endpoint for deleting a single data element

olavsorl:
> **Stacked on #1103.** Base is `refactor/shared-api-key-filter`, which carries `CleanupApiKeyFilter`. Review #1103 first; this PR retargets to `main` automatically once it merges.

## Description

Adds an endpoint for deleting a single data element, its blobs and recording a delete event, callable from inside the cluster with curl and no Altinn token:

```
DELETE storage/api/v1/cleanup/dataelement/{instanceOwnerPartyId}/{instanceGuid}/{dataGuid}
```

```bash
curl -X DELETE -H "X-API-Key: $STORAGE_CLEANUP_API_KEY" \
  "http://localhost:5010/storage/api/v1/cleanup/dataelement/1337/<instanceGuid>/<dataGuid>"
```

It is guarded by a shared secret rather than left unauthenticated, and every call is logged with the deleted element and the caller IP.

The delete goes through `IInstanceMutationRepository.Apply`, so the metadata delete and the delete event land as one transaction and the blobs are cleaned up afterwards via `CleanupDeletedDataElementBlobs` — the same flow `DataController.Delete` uses. No version preconditions are supplied, so a concurrent instance update does not reject an operational delete.

## Instance event actor

`BuildInstanceEvent(eventType, instance, dataElement)` resolves the actor from the claims of the current request and throws when there are none, so an API-key caller would have deleted the element and then failed before the event could be built. It now takes an optional actor and additional info; existing callers pass neither and keep the claims-based resolution unchanged.

The actor recorded is `{ OrgId = instance.Org, AuthenticationLevel = 0 }` with `AdditionalInfo` naming the endpoint. Worth a look during review: that makes an operational delete indistinguishable from an app-owner delete except by the additional info text. A dedicated marker would stand out better, but that is a data-model decision rather than something to slip in here.

## Route identifiers are verified

`PgDataRepository.Read` passes only the data element id to `storage.readdataelement_v2($1)` — the `instanceGuid` argument is unused, so the lookup is global. Without a check, a valid data element belonging to another instance would be deleted against the instance named in the route: the delete event would record the wrong instance, and the blob cleanup would run against the wrong org's storage account. The endpoint now verifies the party id, the instance and the data element's owning instance before deleting anything, and the application is checked for null rather than dereferenced.

## Verification

- `dotnet build` clean, 0 warnings
- `dotnet csharpier check .` clean
- 5 new tests in `CleanupControllerUnitTests`: the happy path (asserting the mutation carries both the delete and the event, and that an explicit actor is supplied), plus one per guard — unknown instance, mismatched party id, data element from another instance, missing application. These are mock-based and run without a database.

## Summary by CodeRabbit

- **New Features**
  - Added secure removal of individual data elements and their associated stored files.
  - Cleanup operations now record instance events with actor information and optional details.
  - Cleanup access is protected by an API key.

- **Bug Fixes**
  - Added validation for instances, parties, data elements, and applications before cleanup.
  - Invalid cleanup requests now return a not-found response without modifying data.
  - Cleanup events identify platform-initiated deletions without attributing them to the application owner.


olavsorl:
@coderabbitai review

Both points from the earlier review should now be addressed:

- The event actor and the instance `LastChangedBy` both record `platform-cleanup` instead of the target application's org, via a `CleanupActor` constant.
- `InstanceEventService.BuildInstanceEvent` is now covered with an explicit actor (asserting the claims fallback is not used), and the cleanup test captures and asserts the additional-information value rather than matching it with `It.IsAny`.

